Issue #36, opened 2017-01-04
Configurable serialization


Flood1993:
As of now, the default method for communication between agents is through pickle serialization. This makes feeding data through outside Python a difficult task. Use of picke should be optional. Add raw message passing?

Peque:
This should be configurable through global/environment variable, per-agent and per-socket as well.

Options: `pickle` (default), `raw`/`none`.

Peque:
JSON, MessagePack, ...?

PR #41, opened 2017-02-02
In reference to issue #36

Flood1993:


osbrain/address.py
```diff
@@ -4,6 +4,7 @@
 from ipaddress import ip_address
 
 import zmq
+import os
```
Peque: Imports from the standard library should go first (i.e.: at the top of the file).

osbrain/address.py
```diff
@@ -171,6 +172,37 @@ def twin(self):
         return self.__class__(self.ZMQ_KIND_TWIN[self])
 
 
+class AgentAddressSerializer():
```
Peque: Maybe it makes more sense to create this class inheriting from `str` (i.e.: have a look at the `AgentAddressTransport` implementation). It simplifies its creation and makes other things easier in the code later. Also, I don't thing the environment default should be loaded here (that should rather be in the Agent's implementation).

osbrain/address.py
```diff
@@ -282,4 +321,6 @@ def twin(self):
         """
         kind = self.kind.twin()
         role = self.role.twin()
-        return self.__class__(self.transport, self.address, kind, role)
+        serializer = self.serializer.serializer_type
+        return self.__class__(self.transport, self.address, kind, role,
+                              serializer)
```
Peque: If you implement `AgentAddressSerializer` as `str`, then you can simply do:

    return self.__class__(self.transport, self.address, kind, role,
                          self.serializer)

osbrain/address.py
```diff
@@ -238,14 +272,17 @@ class AgentAddress():
         Agent kind.
     role : AgentAddressRole
         Agent role.
+    serializer : AgentAddressSerializer
+        Agent serializer.
     """
-    def __init__(self, transport, address, kind, role):
+    def __init__(self, transport, address, kind, role, serializer=None):
```
Peque: I would rather make `serializer` parameter non-optional (default should be loaded in Agent's implementation).

osbrain/agent.py
```diff
@@ -714,15 +716,30 @@ def _process_nonsub_event(self, socket_kind, socket, serialized):
         serialized : bytes
             Data received on the socket.
         """
-        message = pickle.loads(serialized)
+        # TODO Instead of identifying the serializer by looking at the message,
+        #      we should have the AgentAddress corresponding to this channel
+        #      as a parameter, in order to get the information from it.
+        #      This will allow us to get rid of the `try-except` clause.
```
Peque: This TODO should be resolved before merge. :blush:

osbrain/agent.py
```diff
@@ -714,15 +716,30 @@ def _process_nonsub_event(self, socket_kind, socket, serialized):
...
+        # TODO As of now, only `pickle` and `raw` serialization are supported.
```
Peque: This second TODO can be removed (we only support what is implemented in `AgentAddressSerializer` and, for now, only `raw` and `pickle` are expected).

osbrain/agent.py
```diff
@@ -736,9 +753,19 @@ def _process_sub_event(self, socket, serialized):
...
         """
         handlers = self.handler[socket]
-        sepp = serialized.index(b'\x80')
-        data = memoryview(serialized)[sepp:]
-        message = pickle.loads(data)
+        # TODO Instead of identifying the serializer by looking at the message,
+        #      we should have the AgentAddress corresponding to this channel
+        #      as a parameter, in order to get the information from it.
+        #      This will allow us to get rid of the `try-except` clause.
+        # TODO As of now, only `pickle` and `raw` serialization is supported.
```
Peque: About this 2 TODOs: same as just above. :wink:

osbrain/agent.py
```diff
@@ -754,15 +781,51 @@ def _process_sub_event(self, socket, serialized):
     def str2bytes(self, message):
         return message.encode('ascii')
 
+    def _is_address_internal(self, address):
```
Peque: I am not sure this is necessary (i.e.: if we know the `AgentAddress` we should already know about the serialization method).

osbrain/agent.py
```diff
@@ -754,15 +781,51 @@ def _process_sub_event(self, socket, serialized):
...
+                           'inproc://_loopback_safe')
+
     def send(self, address, message, topic=''):
+        """
+        TODO
+        """
         assert isinstance(topic, str), 'Topic must be of `str` type!'
-        serialized = pickle.dumps(message, -1)
+        # Check if socket is for internal use
+        if self._is_address_internal(self.address[address]):
```
Peque: Here again, we should get the information on the serializer type from the `AgentAddress`, not by looking if its "internal" or not (internal sockets should have serialization set as `raw`, so it will work for them as well.

osbrain/agent.py
```diff
@@ -803,8 +866,7 @@ def kill(self):
 
     def close_sockets(self):
         for address in self.socket:
-            if address in ('loopback', '_loopback_safe',
-                           'inproc://loopback', 'inproc://_loopback_safe'):
+            if self._is_address_internal(address):
```
Peque: Not sure if it makes sense to create an `_is_address_internal` method just for this `address in (...)` (assuming it is not used anywhere else).

Peque:
Would you mind adding some docs to this pull request as well? You can create a new file: `docs/source/serialization.rst`.

Thanks a lot for your contributions! :smile:

Peque:
Would you mind rebasing against the latest changes in master? (it will fix some errors you might currently get while running tests).

PR #42, opened 2017-02-03
Serialization - in reference to issue #36

Flood1993:


pytest
```diff
@@ -0,0 +1,6 @@
+reset
+pytest -xv ./ -n 4
+pytest --cache-clear --pep8 -m pep8
+pytest --cache-clear --flakes -m flakes
+pytest --cache-clear --mccabe -m mccabe
+radon cc osbrain/ -s -nb
```
Peque: This file could/should be removed. The user can now easily select what to run without that script.

If you want to run only some tests (having `keyword` in the test name):

    pytest -k keyword

If you want to run just the behavior tests:

    pytest -n 8

If you want to run all tests but for one Python version only (faster):

    tox -e py35

If you want to run all tests for all available Python versions:

    tox

osbrain/tests/test_agent.py
```diff
@@ -229,6 +260,22 @@ def test_pubsub(nsaddr):
...
+    message = b'Hello world'
+    while not a1.get_attr('received'):
+        a0.send('pub', message)
+        time.sleep(0.1)
+    assert a1.get_attr('received') == message
+
+
 def test_agent_inheritance(nsaddr):
     """
     Test agent inheritance; agents can be based on a custom class.
```
Peque: These tests are not well defined. Notice that, if you change `raw` by `pickle`, they all pass. So the functionality could be not implemented without failing (not saying that this is the case).

In order to test the `raw` sockets we could create a single agent and connect with it via a `zmq.Context().socket()` directly to test communications, instead of passing messages between two agents. That is more realistic as the typical use-case for `raw` sockets is when connecting from an external application using "raw" ZMQ.

osbrain/agent.py
```diff
@@ -714,15 +720,24 @@ def _process_nonsub_event(self, socket_kind, socket, serialized):
...
             Data received on the socket.
         """
-        message = pickle.loads(serialized)
+        serializer = self._get_serialization_from_socket(socket)
+
+        if serializer == 'pickle':
+            message = pickle.loads(serialized)
+        elif serializer == 'raw':
+            message = serialized
+
```
Peque: This could be very easily moved to a new method, reducing complexity. In example:
 
    def _deserialize_message(socket, data):
        serializer = self._get_serialization_from_socket(socket)
        if serializer == 'raw':
            return data
        if serializer == 'pickle':
            return pickle.loads(data)
        raise ValueError('Serializer "%s" not supported!' % serializer)

Flood1993:
Lowered the complexity of the functions. Also, the tests for checking message passing without serialization now use raw ZMQ sockets.

However, there is an issue which I have not been able to track down. In order for the tests to run OK, I've had to add explicit calls to close the ZMQ socket and destroy the ZMQ context at the end of each of the following tests (located in osbrain/tests/test_agent.py):
- test_reqrep_without_serialization
- test_pushpull_without_serialization
- test_pubsub_without_serialization

The code that manages the freeing of the resources are the lines:
`socket.close()`
`context.destroy()`

In order to reproduce the issue, first, checkout this commit 33cfa23. Then, run `pytest -k _without_serialization`. The tests should run successfully. If you remove the resource freeing calls from each function, the tests won't pass (I'm getting stuck after passing successfully test_pushpull_without_serialization and test_reqrep_without_serialization, but when running the last one, test_pubsub_without_serialization, it hangs up).

Note that that without the resource freeing calls, each test runs OK individually.

Peque:
Thanks for the changes!

Would you mind...?

- Removing the `pytest` file.
- Moving agent tests to a new file (i.e.: `test_agent_serialization.py`).
- Having a look at why lines 847, 848, 864, 865 are not being covered.

Thanks a lot! :blush:

osbrain/agent.py
```diff
@@ -88,20 +87,16 @@ def __init__(self, name=None, host=None):
...
-        # This in-process socket could, eventually, handle safe access to
-        # memory from other threads (i.e. when using Pyro proxies).
-        socket = self.context.socket(zmq.REP)
-        address = 'inproc://loopback'
-        socket.bind(address)
-        self.register(socket, address, 'loopback', self._handle_loopback)
+        # A loopback socket where, for example, timers are processed
+        self.bind('REP', alias='loopback', addr='loopback',
+                  handler=self._handle_loopback, transport='inproc',
+                  serializer='pickle')
```
Flood1993: Would it be better to set serializer=os.getenv('OSBRAIN_DEFAULT_SERIALIZER')?

Peque: Not really. It is an internal socket (no need to use `raw`) and it passes complex objects/types (such as those passed when using timers) which would be hard to serialize otherwise.

Issue #40, opened 2017-01-24
Test handlers


Peque:
- Test handler with 2 parameters (agent, message)
- Test handler with 3 parameters (agent, message, topic)
- Test topic is properly filtered (no match, partial match, full match)

Peque:
Missing 3 parameter tests (with topic) are tests for PUB-SUB patterns.

PR #43, opened 2017-02-06
Add missing test topic

Flood1993:
This is related to issue #40 

I added some tests to check various cases of the PUBSUB pattern when passing topics around.

The way I'm checking the invalid cases is not very elegant, though. I'm trying to send the message ten times sleeping 0.1 seconds between each try (for a total of 10 tries), and then checking that no message was received.


Peque:
@Flood1993 Thanks for the pull request! :blush:

Do you think maybe the tests could be squashed into one? Would it make sense?

In example: one publisher, which sends messages with different topics during a period of time. And multiple subscribers, each subscribed to different topics. The handler for all subscribers could be to add the received message to a list. Then we can check if all messages are alike or if messages from different topics were received, etc..

Flood1993:
Sounds like a good idea. I will take a look into it 👍

PR #47, opened 2017-02-15
Add tests for the behaviour of raw PUBSUB communication.

Flood1993:
I added some tests for checking the behaviour of the PUBSUB communication pattern when using `raw` serializer. The idea is that it mimics the `raw` ZMQ message passing, in which the topic is sent along with the message, and is responsability of the receiver (in our case, the handler function) to take care of them accordingly.


Peque:
:+1: :blush:

PR #52, opened 2017-03-01
Reply early

Flood1993:
In this PR I'm trying to implement the quick reply option for the REQREP pattern.

The idea behind it is that the handler must have a single `yield` statement (which can only be executed one -no loops-), which will be the reply sent to the requester. Optionally, after the yield, there might be more code to be executed by the server.

I have implemented a test to check the functionality, but it is not passing. Somehow, a variable updated after the yield is not getting modified (or perhaps it is, but due to some other thing, the requester can't see it updated).


Peque:
Maybe we could close this in favor of #54.

PR #54, opened 2017-03-02
Reply early (rebased, fixed and refactored)

Peque:
Rebased #52 against #53. Fixed #52 there.


Flood1993:
Changes look really good! :)

PR #55, opened 2017-03-03
Improve layout

Flood1993:
Simple changes to agent.py so as to make it a little bit more consistent. I'm aware that for lists, list() != []. However, for the dictionary in that case, it seems safe to change the syntax.


osbrain/agent.py
```diff
@@ -304,7 +304,7 @@ def stop_all_timers(self):
         """
         Stop all currently running timers.
         """
-        for alias in list(self._timer.keys()):
+        for alias in self.list_timers():
```
Peque: :+1:

PR #56, opened 2017-03-03
Improve REQ-REP test suite

Peque:
Separate REQ-REP test file, add new tests and cover
`osbrain/tests/common.py` too with tests.

### DIFF
--- a/osbrain/tests/common.py
+++ b/osbrain/tests/common.py
@@ -6,6 +6,25 @@ from osbrain import run_nameserver
 
 
 def logger_received(logger, log_name, message, timeout=1.):
+    """
+    Check if a logger receives a message.
+
+    Parameters
+    ----------
+    logger : Proxy
+        Proxy to the logger.
+    log_name : str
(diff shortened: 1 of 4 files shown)


osbrain/tests/common.py
```diff
@@ -1,7 +1,42 @@
...
+def agent_dies(agent, nsproxy, timeout=1.):
+    t0 = time.time()
+    while True:
+        time.sleep(0.01)
+        if agent not in nsproxy.agents():
+            break
+        if timeout and time.time() - t0 > timeout:
+            return False
+    return True
+
```
Flood1993: Docstrings for these functions seems to be within the tests of those functions, under test_test_common.py. Wouldn't it be better to place it here?

Peque: Yeah, docstrings are missing... shame on me. :joy: I'll add them. ;-)

Peque:
Updated with missing docstrings.

PR #59, opened 2017-03-03
Add first SIGINT test to an agent (not the NameSever agent)

Flood1993:
Added first test related to SIGINTs. As for the first commit, only an agent which is not the NameServer is tested.

In relation to #20

PR #62, opened 2017-03-08
Add context managers for AgentProcess

Flood1993:
Perhaps context managers could help us with the problems we are having in #59 . 

The problem is that as of now, we need to join the process so as for it to get removed from the process table, as it can be seen in `test_agent_shutdown` under `test_agent.py`.

Maybe the use of a context manager could replace at some point the run_agent function.

I'm in the process of further investigating the issue, so do not merge this branch yet, please.

### DIFF
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -1099,6 +1099,16 @@ class AgentProcess(multiprocessing.Process):
         self.sigint = True
         self.kill()
 
+    def __enter__(self):
+        self.start()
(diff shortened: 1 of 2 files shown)


Peque:
I don't think it is very important right now to get this included, so we may postpone this for a 1.0.0+ release (unless it is necessary for #59).

Flood1993:
Closing this since I think it makes no sense (if at all, we could in the future add context managers for agents and not agent processes).

PR #63, opened 2017-03-09
Implement delayed actions

Peque:
A self-contained change that could be added to master while we keep making other changes.

### DIFF
--- a/osbrain/common.py
+++ b/osbrain/common.py
@@ -77,6 +77,41 @@ def repeat(interval, action, *args):
             delay = interval - (time.time() - starttime)
             if event.wait(delay):
                 break
+
     Thread(target=loop).start()
     event.stop = event.set
+
+    return event
+
+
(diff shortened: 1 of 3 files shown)


Flood1993:
Very nice code. Feel free to merge it into master @Peque ! :+1:

osbrain/common.py
```diff
@@ -77,6 +77,41 @@ def loop():
...
+    """
+    event = Event()
+
+    def wait():
+        if event.wait(delay):
+            return
+        action(*args)
+
+    Thread(target=wait).start()
+    event.stop = event.set
```
Flood1993: I'm not sure what it does... Could you explain?

Peque: That line simply allows you to use `timer.stop` instead of `timer.set`, which I think is more intuitive.

Flood1993: Ah, I see :+1:

Issue #65, opened 2017-03-09
Refactor agent sockets


Peque:
Agent code is starting to be a little mess right now. We should refactor sockets to inherit from zmq.sockets (we probably need to subclass zmq.Context too, then).

Sockets should store address information to avoid looking up tables (dictionaries) and to reduce the agent's number of attributes.

PR #66, opened 2017-03-09
Implement asynchronous requests

### LINKED ISSUE #31: Implement async calls
To avoid REQ-REP blocking, which would be useful for non-blocking safe calls, PUB-SUB synchronization, etc..

Peque:
Initial implementation. Should fix #31.

Related: #65

### DIFF
--- a/osbrain/tests/test_agent_async_requests.py
+++ b/osbrain/tests/test_agent_async_requests.py
@@ -92,6 +92,41 @@ def test_yield(nsproxy):
     assert client.get_attr('received') == ['xfoo', 'xbar']
 
 
+def test_unknown(nsproxy):
+    """
+    When receiving a response for an unknow request (or an already processed
+    request), a warning should be logged and handler should not be executed.
+    """
+    def late_reply(agent, request):
+        agent.received.append(request)
+        time.sleep(1)
+        return 'x' + request
+
+    def receive(agent, response):
+        agent.received.append(response)
+
+    server = run_agent('server')
+    client = run_agent('client')
+    logger = run_logger('logger')
+    server.set_attr(received=[])
+    client.set_attr(received=[])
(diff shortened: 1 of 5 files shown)


osbrain/tests/test_agent_async_requests.py
```diff
@@ -0,0 +1,203 @@
...
+    sync_agent_logger(client, logger)
+
+    addr = server.bind('ASYNC_REP', alias='replier', handler=late_reply)
+    client.connect(addr, alias='async', handler=receive)
+
+    # Response received in time
+    client.send('async', 1, wait=2)
+    time.sleep(1.1)
+    assert server.get_attr('received') == [1]
+    assert client.get_attr('received') == ['x1']
```
Flood1993: I'm afraid I don't understand why this assert is passing if the wait is supposed to be 2 seconds and we only sleep 1.1 :S

Peque: Yeah, `wait` is actually a maximum wait delay. It means, if the response is received within `wait` seconds, then everything is okay (this is the case you are referring to, as the response is received after 1 second and we are waiting at most 2 seconds).

See just below for an example in which the response is actually not received in time (we set `wait` to 1 second but the response is received after 2).

Peque:
> And what would you think about trying to reduce the wait time of the timers? So as for when executing the tests locally through tox they run faster?

@Flood1993 Yeah, I've thought about that too. And I have thought about how to rewrite the tests in a more convenient way to monitor changes in agents attributes, for example. I may open an issue soon to discuss it, but I think that would be a post-1.0.0 change. For now, let us keep those wait times just for consistency with other tests.

Flood1993:
Perfect. Everything OK for me. Feel free to go ahead with the merge into master :+1:

Issue #64, opened 2017-03-09
Improve nameserver coverage


Peque:
These two lines are a bit annoying:

https://codecov.io/gh/opensistemas-hub/osbrain/compare/35a869e754ab377db51c831415d2af34e7cb27a2...338b8b488bc9d1a0b6df366550535f3919b1272e/changes

Depending on Travis mood, they might be tested or they might not. We should create a test to ensure we cover those lines. Otherwise Codecov may report an incorrect decrease in test coverage (i.e.: as in #63).

Peque:
Fixed in #69.

PR #67, opened 2017-03-09
Trying to fix test coverage for the two lines

Flood1993:
In relation to #64 .

I tried making a test that creates Agents that shutdown after a certain timeout, which should give `NameServerProcess.shutdown()` enough time to enter the while loop and cover those lines in the test. However, this is failing. My personal guess is that the GIL is somehow related, but I don't know how to get around it.


osbrain/tests/test_nameserver.py
```diff
@@ -77,6 +96,28 @@ def test_nameserver_proxy_shutdown_with_agents():
...
+        def __init__(self, *args, **kwargs):
+            super(NewAgent, self).__init__(*args, **kwargs)
+
+        @Pyro4.oneway
+        def shutdown(self):
+            time.sleep(0.3)
+            super().shutdown()
+
+    run_agent('Agent0', nsaddr=addr, base=NewAgent)
+    run_agent('Agent1', nsaddr=addr, base=NewAgent)
```
Peque: Is this the reason why you implemented `get_nameserver_and_address()` above? If so, can't you just omit the `nsaddr` parameter? (remember to pass the `nsproxy` fixture as parameter to the `test_nameserver_shutdown_with_agents()` function).

Also, if you really needed to get the address of the name server, I think you can do `nsproxy.addr()`. :wink:

Flood1993: Not exactly. I implemented that function since I wanted to directly call `NameServerProcess.shutdown()`. I did some research and couldn't find a way to call it from a NSProxy.

That `get_nameserver_and_address()` function is really a slightly modified `random_nameserver()` (from `nameserver.py` function), main difference being the return value (NameServerProcess itself instead a NSProxy to it).

I have pushed another commit that automatically obtains the address from the NameServerProcess, simplifying the function/test as you suggested :+1: .

Peque:
Already merged #69.

PR #68, opened 2017-03-10
Add test for after calling a method from the parent

Flood1993:
Add one more test for completeness.


Flood1993:
Yes, it seems it was wrong lol. I think we could close this one since in #69 it is specified how this could be done. What do you think @Peque ? Feel free to deny it when you want!

PR #69, opened 2017-03-10
Improve test coverage in NameServerProcess

Peque:
Here is my attempt... Note that if you don't like it we can still consider cherry-picking the first commit (182d3cbb089f2100dd7b4babcf24e3815dea4495), which implements scheduled agent shutdowns (it can be useful anyways).


Peque:
I simplified the code a bit as it didn't make sense to have `agent.shutdown(after=1)` when we could already do `agent.after(1, 'shutdown')`. According to the zen of Python "There should be one-- and preferably only one --obvious way to do it.".

PR #70, opened 2017-03-10
Remove ready method from agent, since it is the same than ping

Flood1993:
I noticed than `Agent.ready()` and `Agent.ping()` were basically doing the same thing.

### DIFF
--- a/osbrain/proxy.py
+++ b/osbrain/proxy.py
@@ -90,7 +90,7 @@ class Proxy(Pyro4.core.Proxy):
             exception.
         """
         try:
-            self.ready()
+            self.unsafe.ping()
         except Exception:
             if time.time() - time0 < timeout:
                 return False
@@ -172,8 +172,7 @@ class Proxy(Pyro4.core.Proxy):
         """
         return (methodname in self._pyroMethods and
                 not methodname.startswith('_') and
-                methodname not in ('ready', 'run', 'get_attr', 'kill',
-                                   'safe_call'))
+                methodname not in ('run', 'get_attr', 'kill', 'safe_call'))
 
     def _remote_call(self, methodname, args, kwargs, flags, objectId):
         """
--- a/osbrain/tests/test_agent.py
+++ b/osbrain/tests/test_agent.py
@@ -50,7 +50,7 @@ def test_early_agent_proxy(nsaddr):
     # Locate agent now
     a0 = Proxy('a0', timeout=3.)
     # Just check agent is ready
-    assert a0.ready() == 'OK'
+    assert a0.unsafe.ping() == 'pong'
 
 
 def test_agent_loopback(nsaddr):
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -400,12 +400,6 @@ class Agent():
         """
         return list(self._timer.keys())
 
-    def ping(self):
-        """
-        A simple ping method (for testing purposes).
-        """
-        return 'pong'
-
     def raise_exception(self):
         """
         Raise an exception (for testing purposes).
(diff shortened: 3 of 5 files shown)


Flood1993:
It makes definetely more sense. I rebased my branch with the changes ^^

osbrain/proxy.py
```diff
@@ -174,7 +174,7 @@ def _is_safe_method(self, methodname):
         """
         return (methodname in self._pyroMethods and
                 not methodname.startswith('_') and
-                methodname not in ('ready', 'run', 'get_attr', 'kill',
+                methodname not in ('ping', 'run', 'get_attr', 'kill',
```
Peque: Watching this I now remember why there were two different methods...

`ready` is expected to be a safe call (i.e.: it does not get serialized to the main thread). Putting `ping` in this line means there are some tests that may not be testing what they should.

I would say: remove `ping` form here. Then instead of replacing the `ready()`s with `ping()`s, try something like `.unsafe.ping()`.

osbrain/proxy.py
```diff
@@ -92,7 +92,9 @@ def _ready_or_timeout(self, time0, timeout):
             exception.
         """
         try:
-            self.ping()
+            safe_status = self._safe
+            self.unsafe.ping()
+            self._safe = safe_status
```
Peque: I don't think you need to save the `._safe` status here (if you do, then it would be a bug). Can you confirm?

osbrain/tests/test_agent.py
```diff
@@ -50,7 +50,9 @@ def test_early_agent_proxy(nsaddr):
     # Locate agent now
     a0 = Proxy('a0', timeout=3.)
     # Just check agent is ready
-    assert a0.ping() == 'pong'
+    safe_status = a0._safe
+    assert a0.unsafe.ping() == 'pong'
+    a0._safe = safe_status
```
Peque: I don't think you need to save the `._safe` status here (if you do, then it would be a bug). Can you confirm?

Flood1993: I answer here so as to not duplicate answers. All tests pass if I remove the "store status" lines, but I think it might not be doing what we want, since the status is then kept even after the call (even though this may be what we want, I'm not sure). Here is a reduced example:

```
class C():
  def __init__(self):
    self.x = 3
    
  @property
  def set_four(self):
    self.x = 4
    return self
    
  def print_x(self):
    print(self.x)
    
c = C()
c.print_x() # prints 3
c.set_four.print_x() # prints 4
c.print_x() # prints 4
```

Since before we were calling ready and ping separately, there was no status modification, but now there is. Perhaps we could create a method for specifically unsafe calling a method, despite configuration being set to safe?

Peque: In that case it would be a bug. I'll have a look at it. :+1:

Peque: I think it is not a bug. See #74.

Peque: Note that `._safe` is a variable that is set to `True`/`False` just after `.safe`/`.unsafe` but then, once the proxy executes one method, its value is set back to `._default_safe`, which is gotten from the environment variable or the parameter when the proxy is initially created.

In your example it would be something like:

```python
class C():
  def __init__(self):
    self.x = 3
    self.default_x = 3

  @property
  def set_four(self):
    self.x = 4
    return self

  def print_x(self):
    print(self.x)
    self.x = self.default_x

c = C()
c.print_x() # prints 3
c.set_four.print_x() # prints 4
c.print_x() # prints 3
```

Peque: As a side note, completely unrelated, you can use:

    ```python
    print('Hello world')
    ```

If you want the code in your comments to get highlighted as Python code.

Flood1993: Oh, I see. So I will remove the status storage from the code ^^. I think we can merge this into master after #74

Peque:
@Flood1993 looks good now. Feel free to merge. :+1:

Consider squashing. :wink:

PR #72, opened 2017-03-13
Code refactored to avoid duplication

Flood1993:
I extracted some code that was duplicated into `_loopback_exception` and `_loopback_reqrep` functions. Also, the `loopback` function has been moved to a different part of the file where it makes more sense due to context.


Peque:
@Flood1993 Looks good to me now. You'll need to rebase again, though. :wink:

PR #75, opened 2017-03-14
Use thread instead of multiplex Pyro server

### LINKED ISSUE #6: Multiplex vs. thread server
The default, currently, is a multiplexed server. This, however, prevents the user from accessing concurrently an agent unless configured to use a threaded server before running it.

In order to allow fine-grained control, a threaded server approach would be better as long as, by default, proxies always used safe calls (serializing calls to the main thread using an inproc socket). With this approach, the user would be able to explicitly change the call to be unsafe and decide, for each call, between safe/unsafe.

Some calls like `send` could be async by default (i.e.: no return value).

Some calls could be changed to be unsafe by default (for example `addr`, or calls that are only reading memory).

- Depends on #30.

Peque:
Initial implementation, still investigating better ways to successfully shutdown in both cases.

Should fix #6.

### DIFF
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -209,8 +209,6 @@ class Agent():
         self.keep_alive = True
         self._shutdown_now = False
         self.running = False
-        # Kill parent AgentProcess
-        self.kill_agent = False
         self._DEBUG = False
 
         self.context = zmq.Context()
@@ -1219,7 +1217,7 @@ class Agent():
 
     def kill(self):
         self.close_sockets()
-        self.kill_agent = True
+        self._pyroDaemon.shutdown()
 
     def close_sockets(self):
         for address in self.socket:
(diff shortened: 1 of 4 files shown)


osbrain/agent.py
```diff
@@ -1219,7 +1217,7 @@ def shutdown(self):
 
     def kill(self):
         self.close_sockets()
-        self.kill_agent = True
+        self._pyroDaemon.shutdown()
```
Flood1993: It works... Why? I can't find where we are setting Agent._pyroDaemon, so my guess is that it works simply because we are always supposed to access the agent/nameserver through a proxy or something like that? And even in that case, I don't understand how it doesn't raise an `AttributeError`. Could you explain a bit please? :)

Peque: Good question.

We are not setting that attribute, but Pyro does:

https://pyro4.readthedocs.io/en/stable/servercode.html#attributes-added-to-pyro-objects

We assume if the `Agent` is running, then it is being served with Pyro (I think it is a sane assumption). If it is not running, then we shouldn't be calling `.kill()` anyway.

Flood1993: Yes, it definetely makes sense. Thanks @Peque ^^

PR #79, opened 2017-03-16
Refactor locate_ns

Flood1993:
Minor changes to the function. I think this way it is slightly more readable. However, feel free to reject it.


Flood1993:
@Peque I realized that would change the behaviour a little bit, from a do-while loop to a simple while loop... I made another commit to restore the original functionality, but it is very similar to the original one. Perhaps we could reject this and leave the original.

Peque:
@Flood1993 Yeah, nice catch! Maybe we should at least always try to locate the name server once (even when the timeout is set to 0...).

Feel free to decide wether to merge changes or not. In my opinion: either leave it as before or, if you are going to make changes, I would simply merge the first commit; we would never find the name server when setting a timeout of 0 (that's bad), but the code is more readable (that's good).

Flood1993:
Okay, I think I will not merge this PR, but I will add some tests on a new PR so as to check precisely this behaviour :+1:

Peque:
That's good!

PR #80, opened 2017-03-17
Add locate_ns timeout tests

Flood1993:
See #79

### DIFF
--- a/osbrain/tests/test_proxy.py
+++ b/osbrain/tests/test_proxy.py
@@ -75,8 +75,27 @@ def test_wrong_nameserver_address():
     """
     Locating a name server that does not exist should raise an error.
     """
+    t0 = time.time()
     with pytest.raises(TimeoutError):
         locate_ns('127.0.0.1:22', timeout=1.)
+    assert 0.5 <= time.time() - t0 <= 1.5
+
+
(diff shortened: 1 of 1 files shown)


osbrain/tests/test_proxy.py
```diff
@@ -79,6 +79,21 @@ def test_wrong_nameserver_address():
         locate_ns('127.0.0.1:22', timeout=1.)
 
 
+def test_no_timeout_locate_ns_not_existing():
+    """
+    Locating a NS that does not exist with no timeout should raise an error.
+    """
+    with pytest.raises(TimeoutError):
+        locate_ns('127.0.0.1:22', timeout=0.)
```
Peque: This could be a "duplicate" of `test_wrong_nameserver_address`. Maybe add something like:

    t0 = time.time()
    with ...:
        locate_ns(...)
    assert time.time() - t0 < 1

?

Peque: Or we could even try `assert ... < 0.5` for that case and then add another assertion for `test_wrong_nameserver_address`: `assert 0.5 < ... < 1.5`.

Flood1993: The second idea is awesome!

Issue #57, opened 2017-03-03
Documentation structure for 1.0.0


Peque:
- ~About~
  - ~Feature overview~
  - ~What can you use osBrain for?~
  - ~Performance~
- ~Introduction~
  - ~Installation~
  - ~Hello world~
  - ~Agents and Proxies~
  - ~The name server~
- ~Basic communication patterns~
  - ~Push pull pattern~
    - ~Example~
    - ~List of handlers~
  - ~Request reply pattern~
    - ~Example~
    - ~Yield vs. return (reply early)~
  - ~Publish subscribe pattern~
    - ~Example~
    - ~Filtering~
- ~Other considerations~
  - ~Clients versus servers~
  - ~Adding new methods~
  - ~Lambdas~
  - ~Shutting down~
  - ~OOP~
- Other patterns? (i.e.: stream)
  - ...
- ~Timers~
  - ~Repeated actions~
  - ~Delayed actions~
  - ~Closing timers (#38, #22)~
- ~Transport protocol (#37, #25)~
  - ~Introduction~
  - ~Global configuration~
  - ~Per-agent configuration~
  - ~Per socket configuration~
- Serialization
  - Introduction
  - Global configuration
  - Per-agent configuration
  - Per socket configuration
- ~Advanced proxy handling~
  - ~Understanding proxies better~
  - ~Executing unsafe calls~
- Distributed systems (#4)
  - Accessing agents from another terminal through proxies
  - Example of distributed systems in the documentation
- ~Security~
- ~Developers~
  - ~Workflow~
  - ~Installing dependencies~
  - ~Running the tests~
  - ~Generating documentation~
- ~API~
- ~Software license and disclaimer~

Peque:
Closed in favor of #156.

PR #88, opened 2017-03-21
Improve documentation/examples regarding serialization

Flood1993:
I have improved the documentation in relation to #57 . 

Also, the examples from the `examples` folder run properly. I have designed them so as to crash when trying to serialize some data which json is not able to serialize, taking into account that de default serializer is pickle, which should be able to serialize it.


docs/source/serialization.rst
```diff
@@ -28,34 +29,59 @@ Specifying the serializer only makes sense in `server` sockets, since `clients`
...
+Specifying the serializer at per agent level will override the global
+configuration. This can be done as follows:
 
-The second one is manually specifying it when creating an agent:
+.. code:: python
 
-.. literalinclude:: ../../examples/explicit_serializer_agent.py
+    ...
+    a1 = run_agent('a1', serializer='json')
+    ...
```
Peque: Maybe you could remove the `...`.

docs/source/serialization.rst
```diff
@@ -28,34 +29,59 @@ Specifying the serializer only makes sense in `server` sockets, since `clients`
...
-- Socket
-- Agent
-- Environment variable.
+.. code:: python
+
+    ...
+    a1 = run_agent('a1', serializer='json')
+    # Raw serialization will override json for this socket
+    addr1 = a1.bind('PUB', 'alias1', serializer='raw')
+    ...
```
Peque: Maybe you could remove the `...`.

Flood1993:
I removed the `...` as suggested, and also thought about squashing all the serialization examples into one, and let the user play around with it in case they wanted to.

PR #94, opened 2017-03-23
SYNC_PUB implementation

Peque:
Just like with the ASYNC_REP implementation, you may want to have a look at the tests (to get an idea of the proposed API) and you may *not* want to have a look at the code... :joy:


osbrain/tests/test_agent_sync_publications.py
```diff
@@ -0,0 +1,238 @@
...
+        time.sleep(0.01)
+    return False
+
+
+def receive(agent, response):
+    agent.received.append(response)
+
+
+def receive_times2(agent, response):
+    agent.received.append(response * 2)
```
Flood1993: Auxiliary function never used?

Flood1993: Well, I added a comment before and now I can't mark it as approved. But definetely the syntax seems really good to me, so feel free to go ahead with the merge  @Peque :+1:

Peque: Thanks, nice catch! :blush:

Issue #84, opened 2017-03-17
Lingering in PUSH-PULL patterns


Peque:
Related to #83.

Add an option to set `linger` in PUSH sockets.

Peque:
This was merged in #96.

@Flood1993 Should we consider setting a default value greater than 0? I was thinking about this and actually, this only comes into play when shutting down (closing agents), so for normal communications we will not break any ZMQ good practices, but when shutting down we may avoid some nasty blocking issues...

Flood1993:
You are right, since it is only important on shutdown, it would make sense to have a default timeout. What would be an appropiate value?

Peque:
@Flood1993 I'd say a value that does not annoy the user too much while waiting... 1 second? Maybe even less. :thinking:

PR #96, opened 2017-03-24
Linger implementation

Flood1993:


Flood1993:
The added tests pass, but some other have been broken. Also, tests for checking the behaviour when fully shutting down the `pusher` should be added (instead of closing the sockets).

Flood1993:
Despite tests passing, they don't pass for the case in which we call `pusher.shutdown()` instead of `pusher.close_sockets()`. I think it is related to the `self._pyroDaemon.shutdown()` call within `kill()`, but I'm not sure.

Flood1993:
Removing the WIP tag. Perhaps we could close this PR and #84 , being aware of the current limitations (read comment above)?

Issue #87, opened 2017-03-21
Configurable HWM


Peque:
Optionally configure HWM behavior (crash vs. wait/drop).

Analyze PUSH-PULL behavior when HWM reached and linger is set.

Flood1993:
I will be updating this comment with my findings.

It seems that the default ZMQ behaviour in the push/pull pattern when the HWM is full on the receiving end, is to simply ignore the new messages.

Peque:
See #97.

PR #97, opened 2017-03-27
Hwm implementation

Flood1993:
My try at enabling the user set the HWM value. Tackles #87 .

Various things to note:
 - I think there is no easy way to modify the drop/block behaviour of the sockets. They seem to be bound to the type of the socket.
 - ~This is currently implemented after #96~
 - I need to make sure that the `test_hwm_agent_blocked()` test is checking what we actually want, and check that it is not passing because of some unexpected error in the background.


Flood1993:
Yes, I think we can forget about this PR :+1:

PR #98, opened 2017-03-28
Implement test for `wait_agent_attr` function

Peque:
I changed the function to be a bit more general, too. Now you can also use it to check that an attribute ends up with a given value (we don't assume it is necessarily a list anymore).

### DIFF
--- a/osbrain/tests/common.py
+++ b/osbrain/tests/common.py
@@ -87,11 +87,49 @@ def agent_dies(agent, nsproxy, timeout=1.):
     return True
 
 
-def wait_agent_list(agent, name='received', length=None, data=None,
+def attribute_match(attribute, length=None, data=None, value=None):
+    """
+    Check if an attribute matches any condition:
+
+    - Minimum length.
+    - Contains an item.
+    - Is exactly a value.
+
+    Parameters
+    ----------
+    attribute : anything
+        The attribute to match against.
+    length : int, default is None
+        Return True if the attribute has this minimum length.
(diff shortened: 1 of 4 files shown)


osbrain/tests/common.py
```diff
@@ -87,11 +87,49 @@ def agent_dies(agent, nsproxy, timeout=1.):
...
+    - Minimum length.
+    - Contains an item.
+    - Is exactly a value.
+
+    Parameters
+    ----------
+    attribute : anything
+        The attribute to match against.
+    length : int, default is None
+        Return True if the attribute has this minimum lenght.
```
Flood1993: ~lenght~ -> length (typo)

osbrain/tests/common.py
```diff
@@ -87,11 +87,49 @@ def agent_dies(agent, nsproxy, timeout=1.):
...
+        Whether the attribute matches any condition.
+    """
+    assert length is not None or data is not None or value is not None, \
+        'No condition passed! will return False always...'
+    if value is not None and attribute == value:
+        return True
+    if length is not None and len(attribute) >= length:
+        return True
+    if data is not None and data in attribute:
+        return True
```
Flood1993: As a nitpicky observation, both in the function parameters and the docstrings, order is `length, data, value`. Could we use the same order here? ^^

Peque: Sure! :blush:

Peque:
Squashed. Will merge once Travis says OK.

Issue #91, opened 2017-03-22
Allow bytes topics


Peque:
Make bytes the default type (topics, channels UUIDs...). Allow `str` types as well, encoding them as UTF always.

Flood1993:
UTF encoding? In the `str2bytes()` and  `bytes2str()` functions we are using ascii. Should we change it to UTF as well?

Peque:
Yes, thanks for pointing that out. :blush:

PR #99, opened 2017-03-28
Various fixes

Flood1993:
Minor improvements I found while refactoring.

One is related to #91 
About the condition, I think it is easier to read in the suggested way.

### DIFF
--- a/osbrain/address.py
+++ b/osbrain/address.py
@@ -274,7 +274,7 @@ class AgentAddress():
     ----------
     transport : str, AgentAddressTransport
         Agent transport protocol.
-    address : int
+    address : str, SocketAddress
         Agent address.
     kind : AgentAddressKind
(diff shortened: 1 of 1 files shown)


osbrain/agent.py
```diff
@@ -771,8 +771,8 @@ def _connect_address(self, server_address, alias=None, handler=None):
         assert server_address.role == 'server', \
             'Incorrect address! A server address must be provided!'
         client_address = server_address.twin()
-        assert not client_address.kind.requires_handler() or \
-            handler is not None, 'This socket requires a handler!'
+        assert not (client_address.kind.requires_handler() and not handler), \
+            'This socket requires a handler!'
```
Peque: I actually prefer something in between:

    assert not client_address.kind.requires_handler() or handler

:joy:

PR #108, opened 2017-03-31
Fix bug affecting osmarkets

Peque:
Now that osbrain uses IPC transport by default, osmarkets was failing because agents did not have a common path to store UNIX socket files. This pull request:

- Implements a common directory for UNIX socket files.
- Improves global configuration (use environment variables as expected)
- Increases minimum required Python version to 3.3 (main distributions already ship more recent Python versions). Python 3.4 will probably be unsupported too in the near future...

### DIFF
--- a/osbrain/__init__.py
+++ b/osbrain/__init__.py
@@ -8,10 +8,13 @@ Pyro4.config.SERVERTYPE = 'thread'
 Pyro4.config.REQUIRE_EXPOSE = False
 Pyro4.config.COMMTIMEOUT = 0.
 Pyro4.config.DETAILED_TRACEBACK = True
-os.environ['OSBRAIN_DEFAULT_TRANSPORT'] = 'ipc'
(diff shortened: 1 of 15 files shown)


osbrain/__init__.py
```diff
@@ -1,4 +1,18 @@
 import os
+import sys
+from pathlib import Path
+
+# While Python 3.4 is supported...
+if sys.version_info < (3, 5):
+    def mkdir(directory, exist_ok=True, parents=True, **kwargs):
```
Flood1993: It seems that only the `directory` parameter is used in the function. What do the others do?

Peque: You are right. I thought about implementing those parameters to create a compliant `.mkdir` method in Python 3.4 `Path().mkdir`. But as we do not require that at the moment, I thought we could always assume `parents=True` and `exist_ok=True`.

I will remove them. Thanks! :blush:

PR #112, opened 2017-04-11
Allow agents and name servers to spawn child processes

Peque:
This implements basic inception capabilities. There is no API defined for explicitly running agents and handling children processes (currently that would be the user's responsibility). That should probably be discussed in a new issue.


Peque:
@Flood1993 It is because the `multiprocessing.Process` class actually [has a `daemon` flag](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Process.daemon) that we were unintentionally overriding... :confused:

That is why spawning new processes was raising as exception, as `.daemon` was not False.

PR #120, opened 2017-05-17
Add on test for various requests handling in ASYNC_PUB

Flood1993:
I added this test because I was thinking that perhaps, in osMarkets, we could have problems when different brains ask router for different history data and he is busy serving one.


osbrain/tests/test_agent_sync_publications.py
```diff
@@ -283,3 +283,41 @@ def test_wait(nsproxy):
     assert wait_agent_attr(client, name='error_log', length=1, timeout=0.5)
     assert server.get_attr('received') == [fast, slow]
     assert 'x' + str(slow) not in client.get_attr('received')
+
+
+def test_various_requests(nsproxy):
```
Peque: I'm not sure this test belongs here. Maybe we are just testing:

- ZeroMQ behavior (i.e.: the poller which processes messages from multiple sockets in a round-robin fashion), which should be tested by `pyzmq` itself
- osBrain general behavior (i.e.: all input messages, by default, are processed sequentially by the main thread), which should be tested elsewhere and is probably tested already.

Once it is done, though, it makes no harm (and maybe it actually tests something related to sync. PUB-SUB that we missed before). :blush:

osbrain/tests/test_agent_sync_publications.py
```diff
@@ -283,3 +283,41 @@ def test_wait(nsproxy):
...
+    client_one.connect(server_addr, alias='sub', handler=receive)
+    client_two.connect(server_addr, alias='sub', handler=receive)
+
+    # Each client makes a request, which should overlap in the server
+    client_one.unsafe.send('sub', delay, handler=receive)
+    time.sleep(0.1)
+    assert server.get_attr('received') == [delay]
+    client_two.unsafe.send('sub', delay, handler=receive)
+    time.sleep(0.1)
+    assert server.get_attr('received') == [delay]  # Still serving client_one
```
Peque: Do we really need to use `.unsafe` here? I don't think so, because the request is asynchronous, but maybe I am missing something (didn't run the test myself).

osbrain/tests/test_agent_sync_publications.py
Peque: I'm not sure this test belongs here. Maybe we are just testing:

- ZeroMQ behavior (i.e.: the poller which processes messages from multiple sockets in a round-robin fashion), which should be tested by `pyzmq` itself
- osBrain general behavior (i.e.: all input messages, by default, are processed sequentially by the main thread), which should be tested elsewhere and is probably tested already.

Once it is done, though, it makes no harm (and maybe it actually tests something related to sync. PUB-SUB that we missed before). :blush:

osbrain/tests/test_agent_sync_publications.py
Peque: Do we really need to use `.unsafe` here? I don't think so, because the request is asynchronous, but maybe I am missing something (didn't run the test myself).

Flood1993: The `.unsafe` wasn't necessary. 😅

osbrain/tests/test_agent_sync_publications.py
Flood1993: You are right, what I'm testing is way too abroad/doesn't add anything really. I'll close this PR. :+1:

PR #129, opened 2017-06-02
Improve time usage in handler tests with wait_agent_attr

Flood1993:


osbrain/tests/test_agent_sync_publications_handlers.py
```diff
@@ -96,8 +94,7 @@ def test_sync_pub_send_handlers(nsproxy, params):
             client.send('sub', 'request!')
```
Peque: Maybe you could check the message that is raised too. In example:

```python
with pytest.raises(...) as error:
    [...]
assert 'Error message' in str(error.value)
```

osbrain/tests/test_agent_sync_publications_handlers.py
```diff
@@ -61,8 +60,7 @@ def test_connect_handler_types(nsproxy, params):
 
     client.connect(addr, alias='sub', handler=handler)
     server.publish()
-    time.sleep(1)
-    assert len(client.get_attr('received')) == 1
+    wait_agent_attr(client, length=1)
```
Peque: Here we have a synchronization problem. We connect the subscriber and then we send a message from the publisher. Although it is very unlikely (because the `connect()` method is kind of slow), there is a chance that the server sends the message when the client is not connected yet... :sweat:

Maybe we need a new function in the `helper` module. `sync_pub_sub_agents` (or something like that) that sends a message from the publisher and expects the client to receive that message. After a timeout it would raise an exception. You could inspire yourself in the `sync_agent_logger`. Note that we are sending messages with different `uuid4()` each to make sure we receive the latest (not one sent before). Only that `sync_agent_logger` function does not have a timeout, but we would like to have one in `sync_pub_sub_agents`.

That would definitely be a new commit (and maybe a new ticket if you can't do it now). :blush:

osbrain/tests/test_agent_sync_publications_handlers.py
Flood1993: Added extra checks in 896b3d0

PR #131, opened 2017-06-02
Handler validation raise instead of assert

Flood1993:
This PR should be looked only after looking at #130, since the commit comes from that one. Extract the handler validation into a function that throws an exception instead of an assert, to guarantee it is always checked.


Flood1993:
Oops, I probably left the "tests failing" in the commit message after rebasing. Fixed already.

Flood1993:
You are right! Rewording and pushing into master!

PR #132, opened 2017-06-13
Add dill serialization support

Peque:


Peque:
@Flood1993 If you agree, I will add this PR and #125 before next release.

Peque:
@Flood1993 As you can see the new serializer is covered by old tests:

https://codecov.io/gh/opensistemas-hub/osbrain/commit/d25313e852fe72b49bfe649c7af82681b400761a

Do you think there are still missing tests anyway? If so, I'll add them before merging, but right now I can't think of a test that is missing for this change.

PR #135, opened 2017-06-14
Improve helper module

Flood1993:
Check all conditions when calling wait_agent_attr.

### DIFF
--- a/osbrain/helper.py
+++ b/osbrain/helper.py
@@ -113,24 +113,26 @@ def agent_dies(agent, nsproxy, timeout=1.):
     return True
 
 
-def attribute_match(attribute, length=None, data=None, value=None):
+def attribute_match_all(attribute, length=None, data=None, value=None):
     """
-    Check if an attribute matches any condition:
+    Check if an attribute matches all of the following specified conditions:
 
     - Minimum length.
     - Contains an item.
     - Is exactly a value.
+
+    Note that only those conditions explicitly passed will be checked.
 
     Parameters
     ----------
     attribute : anything
         The attribute to match against.
     length : int, default is None
-        Return True if the attribute has this minimum length.
+        If specified, the attribute must reach this length (or higher).
     data : anything, default is None
-        Return True if the attribute contains this value.
+        If specified, the attribute must contain this value.
     value : anything, default is None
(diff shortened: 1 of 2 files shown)


osbrain/tests/test_helper.py
```diff
@@ -108,11 +108,11 @@ def test_logger_received_count(agent_logger):
     (42, None, None, 14, False),
     (42, None, None, 42, True),
```
Peque: Would be great to add some tests that actually verify the result when setting multiple attributes. In example:

- length + data: where data is not in the attribute.
- length + data: where length is too short.
- length + data: a positive case

osbrain/helper.py
```diff
@@ -113,69 +113,80 @@ def agent_dies(agent, nsproxy, timeout=1.):
...
     value : anything, default is None
-        Return True if the attribute contains this value.
+        If specified, the attribute must be this value.
 
     Returns
     -------
     bool
         Whether the attribute matches any condition.
     """
     assert length is not None or data is not None or value is not None, \
```
Peque: Also, if you agree, we could remove this assert...

osbrain/tests/test_helper.py
Flood1993: Sure, I will add them now :)

osbrain/helper.py
Flood1993: I think it doesn't harm to have it, just in case someone slips a call without conditions by mistake.

Peque: Note that before it would return False if no conditions where given. That was wrong, hence the assertion. In this case, if no conditions are given, it will return True which, in my opinion, is right. So maybe we don't need the assertion anymore?

Flood1993: I feel like calling the function without conditions would be like placing `assert True` in the code, which might go unnoticed and do not cause any problems but it doesn't seem correct to me... This way we would let the programmer know: "This call you are making is not checking anything!", which I think in any case it will only contribute to better code without any real downfalls. What do you think?

Peque: I think it is not very important so, as you don't agree, we just leave it like that. No problem. :blush:

PR #140, opened 2017-06-16
Refactor agent.py 

Flood1993:
When I was looking as to why the subscription from osmarkets was failing, I discovered two pieces of code that looked very similar and extracted them into `_topics_to_bytes`.

Also, changed `subscribe` to `_subscribe`.

I will squash the commits if approved.

### DIFF
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -24,6 +24,7 @@ import zmq
 from . import config
 from .common import format_exception
 from .common import format_method_exception
+from .common import topics_to_bytes
 from .common import unbound_method
 from .common import validate_handler
 from .common import LogLevel
@@ -867,10 +868,7 @@ class Agent():
         uuid = unique_identifier()
         topic_handlers = {}
         if isinstance(handler, dict):
-            for key, value in handler.items():
-                if isinstance(key, str):
-                    key = key.encode()
-                topic_handlers[channel.uuid + key] = value
+            topic_handlers = topics_to_bytes(handler, uuid=channel.uuid)
         else:
             topic_handlers[channel.uuid] = handler
         topic_handlers[uuid] = self._handle_async_requests
@@ -930,12 +928,8 @@ class Agent():
         """
         if not isinstance(handlers, dict):
             handlers = {'': handlers}
-        # Convert all topics to bytes
-        curated_handlers = {}
-        for topic, value in handlers.items():
-            if isinstance(topic, str):
-                topic = topic.encode()
-            curated_handlers[topic] = value
+
+        curated_handlers = topics_to_bytes(handlers)
         # Subscribe to topics
         for topic in curated_handlers.keys():
             self.socket[alias].setsockopt(zmq.SUBSCRIBE, topic)
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -647,7 +647,7 @@ class Agent():
         self.register(socket, server_address, alias, handler)
         # SUB sockets are a special case
         if kind == 'SUB':
-            self.subscribe(server_address, handler)
+            self._subscribe(server_address, handler)
         return server_address
 
     def _bind_channel(self, kind, alias=None, handler=None, addr=None,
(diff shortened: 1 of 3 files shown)


osbrain/agent.py
```diff
@@ -914,7 +911,27 @@ def _handle_async_requests(self, data):
         else:
             handler(self, response)
 
-    def subscribe(self, alias: str, handlers: Dict[Union[bytes, str], Any]):
+    def _topics_to_bytes(self, handlers: Dict[Union[bytes, str], Any],
```
Peque: Note that this could be a function (does not use `self`).

Peque:
:thinking:

Flood1993:
Extracted method into a function, and moved it to common.py.

osbrain/common.py
```diff
@@ -47,6 +50,33 @@ def format_method_exception(error, method, args, kwargs):
     return type(error)(message)
 
 
+def topics_to_bytes(handlers: Dict[Union[bytes, str], Any], uuid: bytes = b''):
```
Peque: Could you add a test for this function? :blush:

Flood1993:
Added tests and slightly modified the function so as to check passed topics are valid.

osbrain/common.py
```diff
@@ -47,6 +50,36 @@ def format_method_exception(error, method, args, kwargs):
...
+
+    Returns
+    -------
+    Dict[bytes, Any]
+    '''
+    curated_handlers = {}
+
+    for topic, value in handlers.items():
+        if not isinstance(topic, (str, bytes)):
+            raise TypeError('Incorrect topic type: {}'.format(type(topic)))
```
Peque: As we are declaring types in the function footprint, I think we could avoid this manual type-check. Specially taking into account that this function is only used internally.

If you'd rather keep that check, just merge it. :wink:

Flood1993:
Removed manual checks as suggested. Also updated counters with errors.

osbrain/tests/test_common.py
```diff
@@ -30,3 +31,36 @@ def test_loglevel():
...
+        assert k.startswith(b'Utopic_')
+        assert v == 'handler'
+
+    # Wrong parameter (must be a dictionary)
+    with pytest.raises(AttributeError):
+        topics_to_bytes([1, 2])
+
+    # Wrong topic types (must be `str` or `bytes`)
+    with pytest.raises(TypeError):
+        topics_to_bytes({1: 2})
```
Peque: Hehehe. These two `pytest.raises` do still work, but I think we should remove them from the test:

- `AttributeError`: raises because a list does not have `.items()` method.
- `TypeError`: raises because you cannot concat bytes to int.

osbrain/tests/test_common.py
```diff
@@ -30,3 +31,36 @@ def test_loglevel():
     # Invalid initialization
     with pytest.raises(ValueError):
         LogLevel('FOO')
+
+
+def test_topics_to_bytes():
+    """
+    Test topics_to_bytes basic usage and required types.
```
Peque: I would remove the "and required types" from here too.

osbrain/tests/test_common.py
```diff
@@ -30,3 +31,36 @@ def test_loglevel():
...
+        assert v == 'handler'
+
+    # Valid call (with uuid)
+    curated = topics_to_bytes(handlers, uuid=b'U')
+    assert isinstance(curated, dict)
+
+    for k, v in curated.items():
+        assert isinstance(k, bytes)
+        assert k.startswith(b'Utopic_')
+        assert v == 'handler'
```
Peque: Maybe it could be simplified with some:

```python
@pytest.mark.parametrize('uuid', [None, b'U'])
```

To avoid repeating almost the same code.

Otherwise, I would rather split this test into two: one for `str`, another for `bytes`. But I think a parametrization makes more sense in this case.

Flood1993:
Incorporated lastest suggestions. However, for the parametrization, I used `[b'', b'U']` instead of `[None, b'U']`, since the former is the default value of the keyword argument.

PR #143, opened 2017-06-19
Finer topic handling

Flood1993:
Use case: In osMarkets, we want to be able to subscribe to different topics after the `connect` call has been done by a client. Therefore, it will come in handy to be able to easily modify the topics a given SUB socket is subscribed to.

Note that this first implementation is only for direct PUB/SUB topics. I think for SYNC_PUB/SYNC_SUB channels, it will need a bit more work, but the idea would be the same. I will start working on the tests for that case right now, and will push to this PR when they are ready.

### DIFF
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -911,7 +911,7 @@ class Agent():
         '''
         if isinstance(self.address[socket], AgentAddress):
             self.socket[socket].setsockopt(zmq.UNSUBSCRIBE, topic)
-        if isinstance(self.address[socket], AgentChannel):
+        elif isinstance(self.address[socket], AgentChannel):
             channel = self.address[socket]
             socket = channel.receiver
             treated_topic = channel.uuid + topic
@@ -930,7 +930,7 @@ class Agent():
         '''
         if isinstance(self.address[socket], AgentAddress):
             self.socket[socket].setsockopt(zmq.SUBSCRIBE, topic)
-        if isinstance(self.address[socket], AgentChannel):
+        elif isinstance(self.address[socket], AgentChannel):
             channel = self.address[socket]
             socket = channel.receiver
             treated_topic = channel.uuid + topic
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -899,7 +899,7 @@ class Agent():
         return client_address
 
     def unsubscribe_from_topic(self,
-                               socket: Union[AgentAddress, str, zmq.Socket],
+                               socket: Union[AgentAddress, AgentChannel, str],
                                topic: Union[bytes, str]):
         '''
         Unsubscribe a socket from a given topic.
@@ -924,7 +924,8 @@ class Agent():
             treated_topic = channel.uuid + topic
             self.socket[socket].setsockopt(zmq.UNSUBSCRIBE, treated_topic)
 
-    def subscribe_to_topic(self, socket: Union[AgentAddress, str, zmq.Socket],
+    def subscribe_to_topic(self,
+                           socket: Union[AgentAddress, AgentChannel, str],
                            topic: Union[bytes, str]):
         '''
         Subscribe a socket to a given topic.
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -898,7 +898,7 @@ class Agent():
         self.register(socket, register_as, alias, handler)
         return client_address
 
-    def unsubscribe_socket_from_topic(self, socket, topic: bytes):
+    def unsubscribe_from_topic(self, socket, topic: bytes):
         '''
         Unsubscribe a socket from a given topic.
 
@@ -917,7 +917,7 @@ class Agent():
             treated_topic = channel.uuid + topic
             self.socket[socket].setsockopt(zmq.UNSUBSCRIBE, treated_topic)
 
-    def subscribe_socket_to_topic(self, socket, topic: bytes):
+    def subscribe_to_topic(self, socket, topic: bytes):
         '''
         Subscribe a socket to a given topic.
 
@@ -970,7 +970,7 @@ class Agent():
         curated_handlers = topics_to_bytes(handlers)
         # Subscribe to topics
         for topic in curated_handlers.keys():
-            self.subscribe_socket_to_topic(alias, topic)
+            self.subscribe_to_topic(alias, topic)
         # Reset handlers
         self._set_handler(self.socket[alias], curated_handlers)
 
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -898,7 +898,9 @@ class Agent():
         self.register(socket, register_as, alias, handler)
         return client_address
 
-    def unsubscribe_from_topic(self, socket, topic: bytes):
+    def unsubscribe_from_topic(self,
+                               socket: Union[AgentAddress, str, zmq.Socket],
+                               topic: bytes):
         '''
         Unsubscribe a socket from a given topic.
 
@@ -906,6 +908,8 @@ class Agent():
         ----------
         socket
             Identifier of the socket. Must be a valid entry of `self.socket`
+            for the PUB/SUB pattern and a valid entry of `self.address` for
+            the SYNC_PUB/SYNC_SUB pattern
         topic
             topic which we want to unsubscribe from
         '''
@@ -917,7 +921,8 @@ class Agent():
             treated_topic = channel.uuid + topic
             self.socket[socket].setsockopt(zmq.UNSUBSCRIBE, treated_topic)
 
-    def subscribe_to_topic(self, socket, topic: bytes):
+    def subscribe_to_topic(self, socket: Union[AgentAddress, str, zmq.Socket],
+                           topic: bytes):
         '''
         Subscribe a socket to a given topic.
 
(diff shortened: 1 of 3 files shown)


Flood1993:
Related to fixing the problem with the SYNC_SUB subscriptions: The cause was that, subscriptions were working before only because they were being done at `connect(...)` time, in which the channel was passed as a parameter, and therefore, we could access its uuid for prepending it to our topics (since messages sent from the SYNC_PUB socket append the channel uuid always, as a filtering measure).

However, when we wanted to unsubscribe from a topic, we couldn't figure out which was the full name of said topic, because the uuid was not available anywhere anymore. My fix as of now involves copying the `channel.uuid` into its twin, for making it always accesible. Think if there is a better way to do it.

osbrain/agent.py
```diff
@@ -898,6 +898,44 @@ def _connect_and_register(self, client_address, alias=None, handler=None,
         self.register(socket, register_as, alias, handler)
         return client_address
 
+    def unsubscribe_socket_from_topic(self, socket, topic: bytes):
```
Peque: Maybe `unsubscribe_from_topic()` would be better. First parameter could be an AgentAddress, an alias...

Also, if you set the `topic` type (i.e.: `topic: bytest`), do the same for all parameters. :wink:

osbrain/agent.py
```diff
@@ -898,6 +898,44 @@ def _connect_and_register(self, client_address, alias=None, handler=None,
...
+        Parameters
+        ----------
+        socket
+            Identifier of the socket. Must be a valid entry of `self.socket`
+        topic
+            topic which we want to unsubscribe from
+        '''
+        if isinstance(self.address[socket], AgentAddress):
+            self.socket[socket].setsockopt(zmq.UNSUBSCRIBE, topic)
+        if isinstance(self.address[socket], AgentChannel):
```
Peque: Use `if-else` or `if: return; if: return`.

osbrain/agent.py
```diff
@@ -898,6 +898,44 @@ def _connect_and_register(self, client_address, alias=None, handler=None,
...
+        Parameters
+        ----------
+        socket
+            Identifier of the socket. Must be a valid entry of `self.socket`
+        topic
+            topic which we want to subscribe to
+        '''
+        if isinstance(self.address[socket], AgentAddress):
+            self.socket[socket].setsockopt(zmq.SUBSCRIBE, topic)
+        if isinstance(self.address[socket], AgentChannel):
```
Peque: Use `if-else` or `if: return; if: return`.

osbrain/tests/test_agent_sync_publications.py
```diff
@@ -73,6 +73,41 @@ def on_error(agent):
...
+    # By default, client will be subscribed to all topics
+    server.send('pub', 'hello')
+    assert wait_agent_attr(client, name='received', data='hello')
+
+    # Only subscribe to 'TOP' topic
+    client.unsubscribe_socket_from_topic('sub', b'')
+    client.subscribe_socket_to_topic('sub', b'TOP')
+
+    # Message not received since 'TOP' topic not specified in the send call
+    server.send('pub', 'world')
```
Peque: Again we find synchronization problems here.

What do you think about this? We create a function only for tests `sync_pub_sub_append_received` (for example). This function takes a publisher and a subscriber. Assumes the handler in the subscriber is the classic `append_received` (which we use most of the time). Starts publishing messages like increasing natural numbers (in a for loop, not using a timer, we can set a small sleep here). We wait for the first one to be appended in the client. Then we no longer send messages from the publisher. We wait for the client to receive the last number sent from the publisher (once the first is received, all messages after that one should also be received, and we know which number we sent last from the publisher). Then PUB-SUB are definitely connected and in sync. Now we reset `received` in the client and there we go!

What you think?

(This would apply for other tests we have around there, until we implement heart-beating or some other fancy stuff).

Flood1993: I think it is a good idea, however I would still need to set a handler other than `append_received` once the connection is guaranteed to be ok. Perhaps we could do the same that we did in #151 with `subscribe` or some similar implementation?

Peque: Okay, I'm just throwing random ideas... :joy:

Flood1993:
Incorporated the proposed changes.

osbrain/agent.py
```diff
@@ -898,6 +898,51 @@ def _connect_and_register(self, client_address, alias=None, handler=None,
         self.register(socket, register_as, alias, handler)
         return client_address
 
+    def unsubscribe_from_topic(self,
+                               socket: Union[AgentAddress, str, zmq.Socket],
```
Peque: Maybe `Union[AgentAddress, AgentChannel, str]`? `AgentChannel should be there, and I would remove `zmq.Socket`, as the user should never be working with raw sockets in osBrain.

osbrain/agent.py
```diff
@@ -898,6 +898,51 @@ def _connect_and_register(self, client_address, alias=None, handler=None,
...
+        '''
+        if isinstance(self.address[socket], AgentAddress):
+            self.socket[socket].setsockopt(zmq.UNSUBSCRIBE, topic)
+        elif isinstance(self.address[socket], AgentChannel):
+            channel = self.address[socket]
+            socket = channel.receiver
+            treated_topic = channel.uuid + topic
+            self.socket[socket].setsockopt(zmq.UNSUBSCRIBE, treated_topic)
+
+    def subscribe_to_topic(self, socket: Union[AgentAddress, str, zmq.Socket],
```
Peque: Maybe `Union[AgentAddress, AgentChannel, str]`? `AgentChannel should be there, and I would remove `zmq.Socket`, as the user should never be working with raw sockets in osBrain.

osbrain/agent.py
```diff
@@ -898,6 +898,51 @@ def _connect_and_register(self, client_address, alias=None, handler=None,
         self.register(socket, register_as, alias, handler)
         return client_address
 
+    def unsubscribe_from_topic(self,
+                               socket: Union[AgentAddress, str, zmq.Socket],
+                               topic: bytes):
```
Peque: We do allow `Union[bytes, str]` when making subscriptions (see Agent's `_subscribe` method). I think we should do the same here as well. User can directly use `str` (it will automatically be converted to bytes encoded in UTF8, but can use bytes if they prefer to.

osbrain/agent.py
```diff
@@ -898,6 +898,51 @@ def _connect_and_register(self, client_address, alias=None, handler=None,
...
+        if isinstance(self.address[socket], AgentAddress):
+            self.socket[socket].setsockopt(zmq.UNSUBSCRIBE, topic)
+        elif isinstance(self.address[socket], AgentChannel):
+            channel = self.address[socket]
+            socket = channel.receiver
+            treated_topic = channel.uuid + topic
+            self.socket[socket].setsockopt(zmq.UNSUBSCRIBE, treated_topic)
+
+    def subscribe_to_topic(self, socket: Union[AgentAddress, str, zmq.Socket],
+                           topic: bytes):
```
Peque: We do allow `Union[bytes, str]` when making subscriptions (see Agent's `_subscribe` method). I think we should do the same here as well. User can directly use `str` (it will automatically be converted to bytes encoded in UTF8, but can use bytes if they prefer to.

Peque: (maybe this should be considered in the tests as well: test with bytes and test with string).

osbrain/agent.py
Peque: (maybe this should be considered in the tests as well: test with bytes and test with string).

Peque:
@Flood1993 for consistency with how we currently execute the `.connect()` call, I think these methods should be a bit different:

- Unsubscriptions should have the address (or alias), not socket, as first parameter. Same for subscriptions.
- Unsubscriptions should maybe accept a list of topics as second parameters (? not critical, though).
- Subscriptions should follow a closer API to `.connect()`: they should define the handler. This could be a dictionary in the form `{'topic0': handler0, 'topic1': handler1}` just like with `.connect()`. I think this is more important. Note that currently you can do: `agent.connect(..., {'a': handlera, 'b': handlerb})`, then, we you do `agent.subscribe_to_topic(..., topic='c')` (which could probably be renamed to `.subscribe()`), which handler should be use? a or b? I think we should force the user.
- This last point makes me think that maybe we are looking to make `_subscribe()` public, and create an `unsubscribe` function.

What do you think? (We can talk about this tomorrow if you want)

Flood1993:
I think that would be great. Right now the `socket` parameter is not even a socket. However, I think naming it `alias` might be misleading, since it does not have to be an alias. However, I can't come up with a better name (`address_identifier`? Even though it could also be a channel...). I think more functions have this "problem". 

I completely agree with accepting various handlers at the same time through a dictionary, and also with making `_subscribe` public.

Peque:
Maybe `alias` is good. You can only subscribe/unsubscribe already existing addresses/channels, which means that you should already know the alias. Being a public method, it is also more intuitive and perhaps more common to use the alias as well. We should of course accept AgentAddress/Channel as well.

We could also think for a word to use everywhere when we refer to an Agent's address/channel/alias. Which in my head would be simply "address". :confused:

Flood1993:
Closed in favor of #158

PR #148, opened 2017-06-20
Fix typo and test coverage

Flood1993:
The test could ocassionally report a decrease in codecov since the if branch could sometimes only evaluate the `if True:` path.


Peque:
Or "Use `wait_agent_attr` to improve agent test coverage").

Peque:
Ignore the AppVeyor fail (I already changed the configuration to ignore commits which do not have an `appveyor.yaml` file). :sweat_smile:

PR #151, opened 2017-06-27
Function to guarantee PUB/SUB channel in SYNC_PUB

Flood1993:
This function will guarantee the PUB/SUB channel is already up and running after returning, ready for sending data right away without it being dropped (client will be guaranteed to be at the other end of the PUB socket).


Flood1993:
I added two more commits, refactoring the function a bit. I have also been thinking about whether the connection should be done in the `synchronize_sync_pub` function or not, and I came to the conclussion that it should NOT be done there.

Pending things:
- Add tests for the new non-private Agent functions.
- Change the docstring of `synchronize_sync_pub`.
- Change Agent functions `_subscribe` and `UGLY_get_handler` names into more specific ones, and add tests for them.

osbrain/agent.py
```diff
@@ -543,6 +544,15 @@ def register(self, socket, address, alias=None, handler=None):
             self.poller.register(socket, zmq.POLLIN)
             self._set_handler(socket, handler)
 
+    def get_handler(self, alias):
+        """
+        Get the handler associated to a socket given the socket alias.
+
+        Ideally, this should only be called for alias that represent a
+        SUB socket.
+        """
```
Peque: Docstring `Parameters` and `Returns` sections are missing. :wink:

osbrain/agent.py
```diff
@@ -912,7 +922,7 @@ def _handle_async_requests(self, data):
         else:
             handler(self, response)
 
-    def _subscribe(self, alias: str, handlers: Dict[Union[bytes, str], Any]):
+    def subscribe(self, alias: str, handlers: Dict[Union[bytes, str], Any]):
         """
         Subscribe the agent to another agent.
```
Peque: It would be great to complete de docstring here as well now that the method is "user accessible".

osbrain/helper.py
```diff
@@ -192,3 +192,36 @@ def wait_agent_attr(agent, name='received', length=None, data=None, value=None,
...
     return False
+
+
+def synchronize_sync_pub(server, server_alias, client, client_alias):
+    '''
+    Create a SYNC_PUB/SYNC_SUB channel and connect both agents.
+
+    Make sure they have stablished the PUB/SUB communication within the
+    SYNC_PUB/SYNC_SUB channel before returning. This will guarantee that
+    no PUB messages are lost.
```
Peque: Missing "Parameters" section from the docstring.

osbrain/helper.py
```diff
@@ -192,3 +192,36 @@ def wait_agent_attr(agent, name='received', length=None, data=None, value=None,
...
+
+    # Send messages through the PUB socket until the client receives them
+    server.each(0.1, 'send', server_alias, 'Synchronize', alias='_tmp_timer')
+    assert wait_agent_attr(client, name='_tmp_attr', value=True, timeout=5)
+    server.stop_timer('_tmp_timer')
+
+    # Restore the original handler, now that the connection is guaranteed
+    client.subscribe(uuid, original_handler)
+
+    client.del_attr('_tmp_attr')
```
Peque: I am not sure about this implementation. Worst-case there might messages not received by the client when you change the handler. That is really bad. Also, that requires the `.subscribe()` call to be executed again, which I'm not sure is the best thing either.

Would it make sense to ensure synchronization with SYNC-PUB requests? i.e.:

- First connect from the subscriber.
- Then make requests from the subscriber each X seconds until one response is received.
- That is all.

You achieve synchronization while avoiding double `.subcribe()` call and avoid risks of the SUB handler receiving messages sent for the "syncrhonization" process.

You may want to specify an `on_error=lambda x: pass` when sending the request to avoid warning log messages during synchronization.

What do you think?

osbrain/tests/test_agent_pubsub_topics.py
```diff
@@ -143,27 +144,65 @@ def test_pubsub_topics_raw(nsproxy, serializer):
...
+
+    assert message_01 not in a4.get_attr('received')
+    assert b'fooWorld' not in a4.get_attr('received')
+    assert b'foobarFOO' not in a4.get_attr('received')
+    assert b'foBAR' not in a4.get_attr('received')
+
+    assert message_01 not in a5.get_attr('received')
+    assert b'fooWorld' in a5.get_attr('received')
+    assert b'foobarFOO' in a5.get_attr('received')
+    assert b'foBAR' in a5.get_attr('received')
```
Peque: All changes related to receive, received_list, etc. should go in a separate pull-request.

osbrain/helper.py
Flood1993: That seems interesting, I will take a look into it.

osbrain/tests/test_agent_pubsub_topics.py
Flood1993: Yes, I got them a bit messed up 😅 Will fix that before merging (if it is ever about to be merged).

osbrain/helper.py
Flood1993: I think it would not be possible to synchronize them through requests. I am thinking about the case in which a request might change the internal state of the SYNC_PUB agent, and for those cases, an extra hack around would have to be made, I think (changing the handler of the SYNC_PUB momentarily for the requests).

Perhaps what we could do is explain that some messages might be lost in the process or that it should only be used for testing purposes. Now I'm thinking that perhaps the function does not even belong there. In the end, I came up with it to make some tests more consistent.

Any thoughts?

Peque: Hmmmmm.

Yeah, one possibility would be to take that away from there and keep it just for the tests.

Another possibility would be to make all SYNC_PUB sockets implement heart-beating. This is something we should do at some point anyway, hopefully Soon™.

Heart-beating would mean that a SYNC_PUB socket would publish a "heart-beat" (like an "I'm alive" message) periodically, to all subscribed agents. Clients can use this to know if they should reconnect (maybe the server went down).

What we could do for now is to implement just a part of the heart-beating: allow clients to request a heart-beat at any time (i.e.: "are you alive?"). This would be a request from the SYNC_SUB, although it would be treated as a special case and all SYNC_PUB channels should handle it the same way, without taking into account the user-defined handler for other requests. This means adding a "special case" in the protocol, but I don't think it is bad, as requests do not need to be very fast and are infrequent.

We could then use those heart-beat requests from the tests to ensure synchronization.

If you feel like that is too much for now, just take it away from there and open a new issue so we try to improve it for version 1.0.0.

Peque: Maybe heart-beat messages could be published with a special topic (so that clients may decide to subscribe or not to this heart-beat).

Maybe we could use that for now for the tests: from the publisher, we add a timer that publishes heart-beats with a special topic, from the subscriber, when we want to make sure we are fully connected, we just subscribe to this topic and wait for the first heart-beat.

What you think?

Maybe we should anyway open an issue and move this discussion there (we're probably going to discuss these things again when we start working on a good heart-beat implementation).

Flood1993:
I managed to get around this issue in osmarkets without this function. I think we can close this PR and implement the heartbeat system in the future. What do you think @Peque ?

PR #152, opened 2017-06-28
Refactor tests

Flood1993:
A bit of test refactoring. No test functionality should have been modified. Main improvements:
- Reuse the same `receive` function, instead of having it declared in multiple files.
- Refactorization of `test_agent_async_requests` through fixtures.
- Add more consistency in the naming of the tests (`agent.received`, `handler=receive`...)

### DIFF
--- a/osbrain/tests/common.py
+++ b/osbrain/tests/common.py
@@ -6,6 +6,10 @@ from osbrain import run_nameserver
 from osbrain.helper import sync_agent_logger
 
 
+def receive(agent, message, topic=None):
+    agent.received.append(message)
+
+
 @pytest.fixture(scope='function')
 def nsproxy(request):
     ns = run_nameserver()
--- a/osbrain/tests/common.py
+++ b/osbrain/tests/common.py
@@ -10,6 +10,10 @@ def receive(agent, message, topic=None):
     agent.received.append(message)
 
 
+def set_received(agent, message, topic=None):
+    agent.received = message
+
+
 @pytest.fixture(scope='function')
 def nsproxy(request):
     ns = run_nameserver()
--- a/osbrain/tests/common.py
+++ b/osbrain/tests/common.py
@@ -6,7 +6,7 @@ from osbrain import run_nameserver
 from osbrain.helper import sync_agent_logger
 
 
-def receive(agent, message, topic=None):
+def append_received(agent, message, topic=None):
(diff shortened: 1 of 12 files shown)


osbrain/tests/test_agent_async_requests.py
```diff
@@ -55,6 +87,29 @@ def receive(agent, response):
...
+    # Manually remove the pending request before it is received
+    client.set_attr(_pending_requests={})
+    time.sleep(1.1)
+    assert server.get_attr('received') == ['foo']
+    assert client.get_attr('received') == []
+    assert logger_received(logger,
+                           log_name='log_history_warning',
+                           message='Received response for an unknown request!')
+
+
```
Peque: Why move the `test_unknown` position?

osbrain/tests/test_agent_async_requests.py
```diff
@@ -2,39 +2,71 @@
 Test file for asynchronous requests.
 """
 import time
+import pytest
```
Peque: I'd use a blank line to separate built-in packages from external packages (i.e.: between `time` and `pytest`).

This, however, is not really your fault. It should be included in the style/flakes tests. We could add an issue and do it when we find some time. :joy:

osbrain/tests/common.py
```diff
@@ -6,6 +6,10 @@
 from osbrain.helper import sync_agent_logger
 
 
+def receive(agent, message, topic=None):
+    agent.received.append(message)
```
Peque: I like the idea of moving this function to `common`. We could, however, use a more explicit name? In example: `set_received`, `append_received` or something like that. `receive` seems a bit vague.

Maybe even `set_received_handler`, `append_received_handler`...

What you think?

osbrain/tests/test_agent_async_requests.py
Flood1993: I moved it simply to have the test that use the same texture in order.

osbrain/tests/common.py
Flood1993: Adding `_handler` is a bit too much for my taste, but I will change it to `append_received` 👍

Flood1993:
I fixed the whitespace between the imports, renamed the function, extracted set_received as well and removed the position change of the test to clean the git diff. I will add a PR with that change.

PR #158, opened 2017-07-03
Finer topic handling for SUB/SYNC_SUB

Flood1993:
I still need to squash lots of commits and refactor a bit (maybe some parameter names and method signatures).

### DIFF
--- a/docs/source/basic_patterns.rst
+++ b/docs/source/basic_patterns.rst
@@ -176,3 +176,25 @@ subscribed to topic ``a`` only and *Dave* subscribed to topic ``b`` only.
 
 Note how we can specify different handlers for different topics when
 subscribing agents.
+
+.. index:: more_on_filtering
+.. _more_on_filtering:
+
+More on filtering
+-----------------
+
+We can also easily modify the subscriptions at run-time.
+
+In the following example, *Alice* will be publishing messages using topics
+``a`` and ``b`` at the same time. Meanwhile, *Bob* will first subscribe to
+topic ``a``. After a few seconds, he will subscribe to topic ``b`` while
+unsubscribing from topic ``a``.
+
+.. literalinclude:: ../../examples/pub_sub_more_on_filter.py
+
+.. note:: Syntax regarding handlers in the ``subscribe()`` and
+   ``unsubscribe()`` methods is the same that the one used when specifying
+   the handlers on the ``connect()``/``bind()`` call of the SUB socket.
+
+.. warning:: Calls to the ``subscribe()`` method will always override the
+   previous handler for each of the specified topics, if any.
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -1005,6 +1005,9 @@ class Agent():
             treated_topic = channel.twin_uuid + topic
             self.socket[sub_address].setsockopt(zmq.UNSUBSCRIBE, treated_topic)
             del self.handler[self.socket[sub_address]][treated_topic]
+        else:
+            raise NotImplementedError('Unsupported address type %s!' %
+                                      self.address[alias])
 
     def _subscribe_to_topic(self, alias: str, topic: Union[bytes, str]):
         '''
@@ -1023,6 +1026,9 @@ class Agent():
             sub_address = channel.receiver
             treated_topic = channel.uuid + topic
             self.socket[sub_address].setsockopt(zmq.SUBSCRIBE, treated_topic)
+        else:
+            raise NotImplementedError('Unsupported address type %s!' %
+                                      self.address[alias])
 
     def idle(self):
         """
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -25,6 +25,7 @@ import zmq
 from . import config
 from .common import format_exception
 from .common import format_method_exception
+from .common import topic_to_bytes
 from .common import topics_to_bytes
 from .common import unbound_method
 from .common import validate_handler
@@ -556,14 +557,18 @@ class Agent():
             alias = address
         self.socket[alias] = socket
         self.socket[address] = socket
+        self.socket[socket] = socket
         self.address[alias] = address
         self.address[socket] = address
         self.address[address] = address
         if handler is not None:
             self.poller.register(socket, zmq.POLLIN)
-            self._set_handler(socket, handler)
+            if address.kind in ('SUB', 'SYNC_SUB'):
+                self.subscribe(socket, handler)
+            else:
+                self._set_handler(socket, handler)
 
-    def _set_handler(self, socket, handlers):
+    def _set_handler(self, socket, handlers, update=False):
         """
         Set the socket handler(s).
 
@@ -574,7 +579,13 @@ class Agent():
         handlers : function(s)
             Handler(s) for the socket. This can be a list or a dictionary too.
         """
-        self.handler[socket] = self._curated_handlers(handlers)
+        if update:
+            try:
+                self.handler[socket].update(self._curated_handlers(handlers))
+            except KeyError:
+                self.handler[socket] = self._curated_handlers(handlers)
+        else:
+            self.handler[socket] = self._curated_handlers(handlers)
 
     def _curated_handlers(self, handlers):
         if isinstance(handlers, (list, tuple)):
@@ -667,7 +678,7 @@ class Agent():
         self.register(socket, server_address, alias, handler)
         # SUB sockets are a special case
         if kind == 'SUB':
-            self._subscribe(server_address, handler)
+            self.subscribe(server_address, handler)
         return server_address
 
     def _bind_channel(self, kind, alias=None, handler=None, addr=None,
@@ -804,7 +815,7 @@ class Agent():
         if client_address.kind == 'SUB':
             if not alias:
                 alias = client_address
-            self._subscribe(alias, handler)
+            self.subscribe(alias, handler)
         return client_address
 
     def _connect_channel(self, channel, alias=None, handler=None):
@@ -933,9 +944,11 @@ class Agent():
         else:
             handler(self, response)
 
-    def _subscribe(self, alias: str, handlers: Dict[Union[bytes, str], Any]):
+    def subscribe(self, alias: str,
+                  handlers: Dict[Union[bytes, str], Any]) -> None:
         """
-        Subscribe the agent to another agent.
+        Subscribe a SUB/SYNC_SUB socket given by its alias to the given
+        topics, and leave the handlers prepared internally.
 
         Parameters
         ----------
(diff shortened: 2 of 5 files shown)


Flood1993:
As of now, internally, all handlers for SUB sockets are stored as a dictionary, of the form `{topic: handler, ...}`

For the special SUB case, we have to set the socket to listen to that topic, and also store the handler internally so we know what must be called when a message with that topic is received. In bb7e5a8, a bug occurs in which by overriding the handlers, we lose the first topic.

A real life example of that would be osmarkets, in which we might connect to different pairs at any point in time.

See #159

osbrain/tests/test_agent_sync_publications.py
```diff
@@ -73,6 +73,164 @@ def on_error(agent):
...
+    server.each(0.1, 'send', 'pub', 'connecting...')
+    assert wait_agent_attr(client, data='connecting...')
+
+    # Clear subscription to default '' topic (do not receive anything)
+    client.unsubscribe('sub', '')
+
+    # Message not received
+    server.send('pub', 'not received')
+    assert not wait_agent_attr(client, data='not received',
+                               timeout=0.5)
```
Peque: There might be a sync problem in this test I think: maybe we can send from the publisher before the client completed the unsubscription?

Maybe we could use the incremental numbers trick? Publish two topics `a`/`b`, then unsubscribe from one of them. Then wait for the last received value difference (`a` - `b`) to be greater than 1?

Maybe we can create a `wait_for` function, which accepts a function to execute periodically on the agent until it succeeds or the timeout is reached.

osbrain/tests/test_agent_sync_publications.py
```diff
@@ -73,6 +73,164 @@ def on_error(agent):
...
+
+    # Subscribe to two topics
+    client.unsubscribe('sub', '')
+    client.subscribe('sub', handlers={'negate': receive_negate,
+                                      'normal': append_received})
+
+    # Only message from subscribed topic received
+    server.send('pub', 1, topic='negate')
+    server.send('pub', 2, topic='normal')
+    server.send('pub', 'not received', topic='topic_C')
```
Peque: Could we use `.each()` to avoid sync issues in case subscription is not completed?

osbrain/tests/test_agent_sync_publications.py
```diff
@@ -73,6 +73,164 @@ def on_error(agent):
...
+
+    assert wait_agent_attr(client, data=-1)
+    assert wait_agent_attr(client, data=2)
+    assert not wait_agent_attr(client, data=1, timeout=0.5)
+    assert not wait_agent_attr(client, data=-2, timeout=0.5)
+    assert not wait_agent_attr(client, data='not received',
+                               timeout=0.5)
```
Peque: I think the first two `assert not wait` might be redundant. But if you wanted to keep them, maybe set the timeout to `0.5` only for the first `assert not wait`. Then for the rest set it to `0` to make it a bit faster?

osbrain/agent.py
```diff
@@ -930,11 +933,59 @@ def _subscribe(self, alias: str, handlers: Dict[Union[bytes, str], Any]):
...
         # Reset handlers
-        self._set_handler(self.socket[alias], curated_handlers)
+        if isinstance(self.address[alias], AgentChannel):
+            channel = self.address[alias]
+            sub_address = channel.receiver
+            uuid = channel.uuid
+            curated_handlers = topics_to_bytes(handlers, uuid=uuid)
+            self._set_handler(self.socket[sub_address], curated_handlers)
+        else:
+            self._set_handler(self.socket[alias], curated_handlers)
```
Peque: Maybe this is why it fails. It seems we are settings `curated_handlers`, but what you actually want to do is to update this dictionary if it existed before? (not sure, did not run the tests)

Flood1993:
I'm wondering if it would be a good idea to make `_set_handler` a non private method. That way, the user could easily modify the handlers for other communication patterns other than PUB/SUB and SYNC_PUB/SYNC_SUB. What do you think @Peque ?

Flood1993:
I have improved the tests a bit so as to improve their consistency and readability.

osbrain/tests/test_agent_sync_publications.py
```diff
@@ -73,6 +74,152 @@ def on_error(agent):
...
+    # Set up the agents
+    server = run_agent('server')
+    client = run_agent('client', base=Client)
+
+    server.set_attr(count=1)
+    addr = server.bind(socket_type, alias='pub', handler=lambda: None)
+    client.connect(addr, alias='sub', handler={'odd': append_received,
+                                               'even': append_received})
+
+    # Make sure the connection is stablished
```
Peque: "established"?

osbrain/tests/test_agent_sync_publications.py
```diff
@@ -73,6 +74,152 @@ def on_error(agent):
...
+
+    assert wait_agent_attr(client, data=-1)
+    assert wait_agent_attr(client, data=2)
+    assert not wait_agent_attr(client, data=1, timeout=0.5)
+    assert not wait_agent_attr(client, data=-2, timeout=0.0)
+    assert not wait_agent_attr(client, data='not received', timeout=0.0)
+
+
+@pytest.mark.parametrize('socket_type', ['PUB', 'SYNC_PUB'])
+def test_resubscribe(nsproxy, socket_type):
```
Peque: Not sure if this `test_resubscribe` is an interesting/required test. Even if so, I would call it `test_unsubscribe_and_subscribe_again`.

For me, a `test_resubscribe` would be one that actually checked the behavior in this case:

```python
client.subscribe('sub', handler={'x': foo})
client.subscribe('sub', handler={'x': bar})
```

Subscription after subscription, without an `unsubscribe` call in between.

This is currently not defined behavior. In my opinion we could simply override the first subscription. What you think?

osbrain/agent.py
```diff
@@ -933,9 +944,11 @@ def _handle_async_requests(self, data):
         else:
             handler(self, response)
 
-    def _subscribe(self, alias: str, handlers: Dict[Union[bytes, str], Any]):
+    def subscribe(self, alias: str,
+                  handlers: Dict[Union[bytes, str], Any]) -> None:
```
Peque: I think we should use `handler` (singular) to be consistent with `bind`/`connect`. What you think?

Flood1993: I agree. Changed all occurrences!

osbrain/tests/test_agent_sync_publications.py
Flood1993: 😅

osbrain/tests/test_agent_sync_publications.py
```diff
@@ -73,6 +74,152 @@ def on_error(agent):
...
+
+    # Make sure the connection is stablished
+    server.each(0.1, 'send', 'pub', 'connecting...', topic='odd', alias='tmp')
+    assert wait_agent_attr(client, data='connecting...')
+    server.stop_timer('tmp')
+
+    # Wait until client receives some numbers
+    client.set_attr(received=[])
+    server.each(0.1, publish)
+    assert wait_agent_attr(client, length=5)
```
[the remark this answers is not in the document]
Flood1993: Yes, maybe it would be easier to read and understand the test. Maybe we can change it towards the end of the PR? To get the most important things tackled first.

osbrain/tests/test_agent_sync_publications.py
Flood1993: I agree with everything you said. I renamed the test, and I will work now on writing the true ``test_resubscribe``. I will comment again when I'm done with it. 

For me, I would leave the ``test_unsubscribe_and_subscribe_again`` in the test code base, as it won't hurt to have more tests. What do you think?

Flood1993: Implemented the new test - see commit 9d44e10.

Peque: Yeah, we can leave it there. :blush:

Flood1993:
I pushed various commits with the following changes:
- Remove specifying the topics as bytes. You are right, it makes more sense to use strings plus the use of bytes is undocumented.
- I changed the tests to implement the new functionality you proposed, checking the tail of the messages.
- From the last point, I refactored the server agents to share the same timer for their publications, so as to try to guarantee that the tails are always ``...1, 2]``, or at least for most of the time, because it gave me problems keeping the timers separated.
- I also changed some strings from the test. I feel many of them are still meaningless, I feel like I should try to make the tests more readable for future maintenance, but I'm not sure if the proper way is with in-line comments or in the docstring of the test.

Let me know what you think @Peque so we can keep working on this PR 😄

Flood1993:
Uhh, I don't know why it's crashing on Travis (at least I'm able to reproduce the crashes in my Ubuntu VM). In Windows all tests are passing OK. I will investigate and come back when I find something.

osbrain/tests/test_agent_sync_publications.py
```diff
@@ -59,6 +60,13 @@ def reply(self, delay):
         return 'x' + str(delay)
 
 
+class ServerNumbers(Agent):
+    def publish(self):
+        self.each(0.1, 'send', 'pub', 1, topic='negate')
+        self.each(0.1, 'send', 'pub', 2, topic='normal')
+        self.each(0.1, 'send', 'pub', 'not received', topic='topic_C')
+
```
Peque: If you want to make sure messages are sent in a given order, why don't you do:

```python
def publish(self):
    self.send('pub', 1, topic='negate')
    self.send('pub', 2, topic='normal')
    self.send('pub', 'not received', topic='topic_C')

# [...]
each(0.1, 'publish')
```

?

Just started from here my review, so maybe I'll find later something that makes me understand this better... :sweat_smile:

osbrain/tests/test_agent_sync_publications.py
```diff
@@ -73,6 +81,175 @@ def on_error(agent):
     agent.error_log.append('error')
 
 
+def match_tail(agent, length, data):
+    return agent.received[-length:] == data
```
Peque: Or even...

```python
def match_tail(agent, data):
￼     return agent.received[-len(data):] == data
```

?

osbrain/tests/test_agent_sync_publications.py
```diff
@@ -73,6 +81,175 @@ def on_error(agent):
...
+    # Wait until client receives some numbers
+    client.set_attr(received=[])
+    server.each(0.1, publish)
+    assert wait_agent_attr(client, length=5)
+
+    # Unsubscribe from odd numbers (receiving only even numbers)
+    client.unsubscribe('sub', 'odd')
+
+    assert wait_agent_condition(client, check_non_consecutive)
+    assert client.get_attr('received')[-1] % 2 == 0
```
Peque: Looks good. I would, however, check that `assert not client.execute_as_method(check_non_consecutive)` before unsubscribing. Just to make sure we were receiving consecutive numbers before and that we are not now.

osbrain/tests/test_agent_sync_publications.py
```diff
@@ -73,6 +81,175 @@ def on_error(agent):
...
+    server.each(0.1, 'publish')
+
+    assert wait_agent_condition(client, match_tail, 8, [-1, 2] * 4)
+
+    # Unsubscribe and reset received messages
+    client.unsubscribe('sub', ['negate', 'normal'])
+    client.set_attr(received=[])
+
+    # Nothing should be received (length == 0)
+    assert not wait_agent_attr(client, length=1, timeout=0.5)
```
Peque: Why don't you unsubscribe from one of the topics only so that you can do:

```python
assert wait_agent_attr(client, match_tail, [-1] * 8)
```

instead?

That means, positive assertion, with no need for a timeout (less risk). You can decrease the publisher period to `0.01` to make things faster.

Also, this way you do not need to reset the `received` attribute. :wink:

Peque:
@Flood1993 I just suggested you exactly that (it looked suspicious). :joy:

osbrain/tests/test_agent_sync_publications.py
Flood1993: At first I did what you proposed, but then the code should be ``client.unsubscribe('sub', 'normal')`` 😅 . So I changed it for the other way around, just to avoid any subliminal messages hahaha.

docs/source/basic_patterns.rst
```diff
@@ -176,3 +176,29 @@ subscribed to topic ``a`` only and *Dave* subscribed to topic ``b`` only.
...
+unsubscribing from topic ``a``.
+
+.. literalinclude:: ../../examples/pub_sub_more_on_filter.py
+
+.. note:: Syntax regarding handlers in the ``subscribe()`` and
+   ``unsubscribe()`` methods is the same that the one used when specifying
+   the handlers on the ``connect()``/``bind()`` call of the SUB socket.
+
+.. note:: Calls to the ``subscribe()`` method will always override the
+   previous handler for each of the specified topics, if any.
```
Peque: Maybe this could be a warning?

docs/source/basic_patterns.rst
```diff
@@ -176,3 +176,29 @@ subscribed to topic ``a`` only and *Dave* subscribed to topic ``b`` only.
...
+.. _more_on_filtering:
+
+More on filtering
+-----------------
+
+We have seen how to filter messages on agents at the subscriber side of the
+communication.
+
+It is also remarkable that we can easily modify the subscriptions at runtime
+easily.
```
Peque: Maybe we can simplify this two paragraphs and say it like: "We can also very easily modify the subscriptions at run-time." (says the same with less words?).

osbrain/agent.py
```diff
@@ -933,29 +944,82 @@ def _handle_async_requests(self, data):
...
+
+    def unsubscribe(self, alias: str, topic: Union[bytes, str]) -> None:
+        '''
+        Unsubscribe a SUB/SYNC_SUB socket given by its alias from a given
+        specific topic, and delete its entry from the handlers dictionary.
+        '''
+        if isinstance(topic, (tuple, list, dict)):
+            for t in topic:
+                self.unsubscribe(alias, t)
+            return
```
Peque: This is not currently tested. Could you add a test? And also maybe update the docstring to say that a list/tuple is accepted in order to unsubscribe from multiple topics.

osbrain/tests/test_agent_sync_publications.py
```diff
@@ -73,6 +81,173 @@ def on_error(agent):
...
+
+    Two non consecutive numbers must be received consecutively at that point,
+    since they both will be even.
+    '''
+    def check_non_consecutive(agent):
+        """
+        Check whether the last two received numbers are non-consecutive.
+        """
+        if len(agent.received) < 2:
+            return False
```
Peque: This (`if len(agent.received) < 2`) seems to never happen. If it does not, then remove it. Otherwise make sure it always happens to ensure maximum coverage of our tests as well. :wink:

Peque:
@Flood1993 Pushed one commit here to avoid setting the channel's UUID with a non-unique identifier (it kind of violates the whole purpose of a unique identifier :joy:).

Two tests fail due to the "shutdown many agents test". That was expected after the last shutdown modification. See #208.

Apart from that there are only a couple of minor comments from yesterday and this should be ready for merge. :blush:

Peque:
@Flood1993 A fix for the "shutdown many agents test" is read in #209. Seems to work fine.

Flood1993:
@Peque I did the changes to the documentation and also added a test for multiple unsubscriptions in the same call. I removed the option to pass the topics as a dictionary since it seems counter intuitive to me. Do you think I should leave it there?

Peque:
@Flood1993 Seems `test_agent_sync_publications.py` coverage was reduced. You can correct those lambdas. Try with:

```python
addr = server.bind(..., lambda agent, message: None)
```

Or maybe even better, use an existing handler:

```python
addr = server.bind(..., handler=append_receive)
```

Peque:
@Flood1993 There are also two lines not branch-covered in `agent.py`:

screenshot from 2017-11-10 12-34-44

Could you have a look at that? I would prefer to merge only code that either keeps or improves code coverage, but never decreases it (unless very special circumstances).

PR #160, opened 2017-07-10
Allow attributes to be set when running an agent

Peque:
It is a common task, so this makes it easier for the end user.

In our walk-forward analysis system it also allows to take advantage of the COW on Unix systems and reduce memory usage (we set big read-only market data history dataframes from the parent process using this new parameter and we avoid complete copies on each backtest worker).

### DIFF
--- a/docs/source/considerations.rst
+++ b/docs/source/considerations.rst
@@ -108,3 +108,27 @@ however you may notice some differences:
    to. As the binding was executed on initialization, we need to use the
    ``addr()`` method, which will return the address associated to the alias
    passed as parameter (in the example above it is ``main``).
+
+
+Setting initial attributes
+==========================
+
+Many times, after spawning an agent, we want to set some attributes, which
+may be used to configure the agent before it starts working with the rest of
+the multi-agent system:
+
+.. code-block:: python
+
+   a0 = run_agent('foo')
+   a0.set_attr(x=1, y=2)
+
+It is such a common task that a parameter ``attributes`` can be used when
+running the agent for exactly that:
+
+.. code-block:: python
+
+   a0 = run_agent('foo', attributes=dict(x=1, y=2))
+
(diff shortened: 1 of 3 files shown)


Peque:
@Flood1993 I don't think that is necessary right now. Do you like the `init` name for the attribute? (that was the hardest part of this PR... :joy:)

Flood1993:
It is hard to come up with a better name, to be honest :P. Should we sacrifice some extra characters and name it `attributes`? 

Just for comparison:
`a0 = run_agent('foo', init=dict(x=1, y=2))`
`a0 = run_agent('foo', attributes=dict(x=1, y=2))`

docs/source/considerations.rst
```diff
@@ -108,3 +108,27 @@ however you may notice some differences:
...
+
+   a0 = run_agent('foo')
+   a0.set_attr(x=1, y=2)
+
+It is such a common task that a parameter ``init`` can be used when running
+the agent for exactly that:
+
+.. code-block:: python
+
+   a0 = run_agent('foo', init=dict(x=1, y=2))
```
Flood1993: Now that I think about it, do you prefer that syntax? I personally prefer `init={'x': 1, 'y': 2}`, but its up to you 👍

Peque: It is a bit more compact and perhaps looks more similar to standard parameter values (i.e.: `function(x=1, y=2)`. Anyway the user is free to use their preferred way. When we don't agree or there is not a good reason, better leave it as-is. In this case it is just a matter of preference (I also many times prefer the use of `{}`). :stuck_out_tongue_winking_eye:

Peque:
@Flood1993 Yeah, that is the problem, I also thought about `attributes`. Maybe it is a better name... I don't know. :sweat_smile:

Peque:
@Flood1993 changed name to `attributes`. :wink:

PR #163, opened 2017-07-11
Use `logger_received()` to avoid uncovered sleeps

Peque:
Should avoid this coverage regression:

https://codecov.io/gh/opensistemas-hub/osbrain/pull/160/src/osbrain/tests/test_logging.py

### DIFF
--- a/osbrain/tests/test_logging.py
+++ b/osbrain/tests/test_logging.py
@@ -2,11 +2,11 @@
 Logging module tests.
 """
 import os
-import time
 
 from osbrain import run_agent
 from osbrain import run_logger
 from osbrain.logging import pyro_log
+from osbrain.helper import logger_received
 
 from common import nsproxy  # pragma: no flakes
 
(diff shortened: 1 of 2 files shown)


osbrain/tests/test_logging.py
```diff
@@ -24,26 +24,22 @@ def test_logging_level(nsproxy):
...
+    # Log values
+    assert logger_received(logger, 'some information', 'log_history_info', -1)
+    assert logger_received(logger, 'some warning', 'log_history_warning', -1)
+    assert logger_received(logger, 'some error', 'log_history_error', -1)
+    assert logger_received(logger, 'some debug', 'log_history_debug', -1)
+    # Log sizes
     assert len(logger.get_attr('log_history_info')) == 1 + before
     assert len(logger.get_attr('log_history_warning')) == 1
     assert len(logger.get_attr('log_history_error')) == 1
     assert len(logger.get_attr('log_history_debug')) == 1
```
Flood1993: Should we check that (pseudo code) `len(log_history) == sum(len(log) for log in other_logs)`?

Peque: I would rather have separate assertions so that the errors (if they ever occur) are more clear. In this case it will not save many lines (only the last 3 are expected to have length 1).

Flood1993: Yes, I would leave those assertions as they are, I meant to add one specifically checking `Logger.log_history` (which I think is supposed to contain all logs in order).

Peque: Oh, didn't understand you well. Okay, I'll add it (that's more than refactoring though). :stuck_out_tongue_winking_eye:

Peque: I have finally left that last change out and kept only the code refactoring. :innocent: (squashed and waiting for Travis to say okay)

Flood1993: Perfect ^^

Issue #165, opened 2017-07-11
Shutting down many agents


Peque:
When a name server tries to shut down many agents, they all make proxies to the name server and the threaded Pyro4 pool gets saturated.

This test reproduces the error:

```python
def test_nameserver_proxy_shutdown_with_many_agents():
    """
    Shutdown a name server from a name server proxy when there are many agents
    registered in the name server (make sure proxies do not saturate the name
    server on shutdown).
    """
    import Pyro4
    Pyro4.config.THREADPOOL_SIZE = 4
    ns = run_nameserver()
    for i in range(20):
        run_agent('Agent%s' % i)
    ns.shutdown()
```

This bug affects our walk-forward exploration architecture.

Peque:
Opened a pull request: #166 (an actual fix is currently missing)

Peque:
Fixed in #166.

PR #166, opened 2017-07-11
Add test to reproduce many agents shutdown bug

Peque:
See #165.


Flood1993:
Note: This test passes on Windows, even when setting the number of agents created to 50.

PR #170, opened 2017-07-13
Shutdown bug

Peque:
Fixes #137. Hopefully...

The relevant commit is the "Kill non-running agents with one-way calls". The other commits are just some code clean-ups/fixes that I found while implementing this.

PR #174, opened 2017-07-21
Wait agent condition

Peque:
See #170 first.

Implement a new helper function `wait_agent_condition` which should be useful for tests.

Improve coverage a bit.

### DIFF
--- a/osbrain/helper.py
+++ b/osbrain/helper.py
@@ -193,3 +193,33 @@ def wait_agent_attr(agent, name='received', length=None, data=None, value=None,
             break
         time.sleep(0.01)
     return False
+
+
+def wait_agent_condition(agent, condition, *args, timeout=3, **kwargs):
+    """
+    Wait for an agent's condition to be true.
+
+    The condition is passed as a function which must take, at least, the
+    actual agent as a parameter, and must evaluate to a boolean result.
+
+    Parameters
+    ----------
+    agent : Proxy
+        A proxy to the agent.
+    condition : Callable
+        A function that evaluates the desired condition.
+    timeout : float, default is 3
+        After this number of seconds the function will return `False`.
+
+    Returns
+    -------
+    bool
+        Whether the specified condition was True.
+    """
+    t0 = time.time()
+    while True:
+        if agent.execute_as_method(condition, *args, **kwargs):
+            return True
+        if time.time() - t0 > timeout:
+            return False
+        time.sleep(0.01)
--- a/osbrain/tests/test_bugs.py
+++ b/osbrain/tests/test_bugs.py
@@ -2,9 +2,9 @@
 Test file for bugs found in osbrain.
 """
 import sys
-import time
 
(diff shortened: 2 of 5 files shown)


osbrain/helper.py
```diff
@@ -193,3 +193,33 @@ def wait_agent_attr(agent, name='received', length=None, data=None, value=None,
...
         time.sleep(0.01)
     return False
+
+
+def wait_agent_condition(agent, condition, *args, timeout=3, **kwargs):
+    """
+    Wait for an agent's condition to be true.
+
+    This condition is passed as a function whose only parameter is the actual
+    agent.
```
Flood1993: `whose only parameter is ...` seems to be wrong (since we can pass more parameters around). What about something like: `The condition is passed as a function which must take, at least, the actual agent as a parameter, and must evaluate to a boolean result`.

osbrain/tests/test_bugs.py
```diff
@@ -21,10 +21,6 @@ def inc(agent):
...
-    while True:
-        count = agent.get_attr('count')
-        if count > sys.getrecursionlimit():
-            return
-        time.sleep(0.1)
-        if count == agent.get_attr('count'):
-            raise RuntimeError('Probable recursion level exceeded!')
+
+    limit = sys.getrecursionlimit()
+    assert wait_agent_condition(agent, lambda agent: agent.count > limit)
```
Flood1993: Note that before, there was no time limit for this test, but now the time limit is set to 3 seconds by the `wait_agent_condition(...)` call.

In my case, the recursion limit is 1000 but in Windows, it only seems to be able to get to around 700. Perhaps we could increase the timer for this particular test.

Peque: Your Windows is slower than Travis... :joy:

What if you change the `time.sleep` to `0.05` instead of `0.01`? Does it change? We could increase the delay instead or as well.

Flood1993: Do you mean the `time.sleep(...)` call in `wait_agent_condition(...)`? I think those sleeps should have no effect in this case, even though changing the value does not make the test pass (didn't check the iterations reached).

Peque: Yeah, I meant that sleep. Do you know why is it so slow in Windows?

Flood1993: I have no idea, I will try to figure it out... As of now, I can only confirm that in Windows, that test from the current master branch takes around 4.5 seconds, while it is taking around 0.9 seconds in Ubuntu in my VM, even though it has assigned very few resources...

Peque:
@Flood1993 Changed the timeout to 10 seconds. Tell me if that fixes the problem.

Flood1993:
Yes, now the test passed (taking 5.48 seconds). Since we are exiting early in many of the helper functions when things go as expected, we could even increase those waiting timers for the tests by default. What do you think?

Issue #162, opened 2017-07-10
Shutdown by default on examples


Peque:
Right now the initial examples do not use the `shutdown` method (it is explained later in the documentation).

Maybe we should introduce the shutdown method from the very beginning, as this way the examples can be run without the need to be interrupted with CTRL+C (they do not block forever). This might be more intuitive for beginners than having to kill all the system "by hand".

@Flood1993 what you think?

Flood1993:
I think it would be a great addition ^^

Flood1993:
Solved in #183

PR #183, opened 2017-07-26
Shutdown examples

Flood1993:
All examples tested and working except `dispatcher.py` and `ventilator.py`, they are crashing with error: `NameError: name 'time' is not defined`, and I don't understand why 😕 I am unable to reproduce the issue on Ubuntu, only crashes on Windows 10.

Review those two.

Tackles #162 

Edit: On ``dispatcher.py``, adding a ``time.sleep(4)`` call right after the ``import time`` does not crash at that point, even though it crashes with the same error afterwards.

### DIFF
--- a/docs/source/considerations.rst
+++ b/docs/source/considerations.rst
@@ -93,17 +93,22 @@ In fact, the only difference is the binding from Alice, in which we are using
 a lambda function for the handler.
 
 
+.. _shutting_down:
+
 Shutting down
 =============
 
-Although not covered in the examples until now (because many times you just
-want the multi-agent system to run forever until, perhaps, an event occurs),
-it is possible to actively kill the system using proxies:
+If we want to end the execution of a specific agent in our system, we can do
+it by calling the `agent.shutdown` method:
 
 .. literalinclude:: ../../examples/shutdown.py
 
-.. note:: Shutting down the nameserver will result in all agents registered in
-   the name server being shut down as well.
+Shutting down a name server will result in all agents registered in that name
+server being shut down as well. This allows us to easily shutdown groups of
+agents at the same time.
+
+.. note:: We can stablish connections between agents registered in different
+   name servers.
 
 
 OOP
--- a/docs/source/introduction.rst
+++ b/docs/source/introduction.rst
@@ -44,6 +44,11 @@ Running this example from your terminal should simply show you a log message
 saying `Hello world!` but, what exactly is happening there?
 
 
+.. note:: The ``ns.shutdown()`` calls in the examples are there because we do
+   not want examples to run indefinetely. For more information on that method,
+   refer to the :ref:`shutting_down` section.
+
+
 .. index:: agents, proxies
 .. _agents_and_proxies:
 
--- a/examples/dispatcher.py
+++ b/examples/dispatcher.py
@@ -13,6 +13,7 @@ def request_work(agent):
     x = agent.send_recv('dispatcher', 'READY!')
     if not x:
         agent.shutdown()
+        return
     time.sleep(x)
     agent.send('results', '%s finished with %s' % (agent.name, x))
 
@@ -33,3 +34,8 @@ if __name__ == '__main__':
         worker.connect(results_addr, alias='results')
         worker.connect(dispatcher_addr, alias='dispatcher')
         worker.each(0., request_work)
+
+    while len(ns.agents()) > 2:
+        time.sleep(0.1)
+
+    ns.shutdown()
--- a/examples/various_nameservers.py
+++ b/examples/various_nameservers.py
@@ -1,0 +1,50 @@
+"""
+This example is relatively complex.
+
+We will be creating a system in which agents will connect to different
+name servers. Each name server will therefore represent a 'group' of agents.
+That way, we can easily shut down all agents belonging to a group, without
+interfering with the others.
+"""
+import time
+
+from osbrain import run_agent
+from osbrain import run_nameserver
+
+
+def annoy(agent, message):
+    agent.send('annoy', message)
+
+
+def log_message(agent, message):
+    agent.log_info('Received: %s' % message)
+
+
+if __name__ == '__main__':
+
+    ns = run_nameserver()
+    ns_annoy = run_nameserver()
+
+    # Create normal PUB/SUB communication
+    listener = run_agent('listener', nsaddr=ns.addr())
+    addr = listener.bind('SUB', alias='sub', handler=log_message)
+
+    speaker = run_agent('speaker', nsaddr=ns.addr())
+    speaker.connect(addr, alias='annoy')
+    speaker.each(0.2, annoy, 'Blah blah...')
+
+    # Create annoying agents registered in another name server
+    annoyer_apple = run_agent('annoyer0', nsaddr=ns_annoy.addr())
+    annoyer_apple.connect(addr, alias='annoy')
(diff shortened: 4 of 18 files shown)


Peque:
BTW: for now Windows is not an officially supported platform, so I would forget about those errors in Windows. Anyway... did those examples work well before these changes? :-/

We may want to add the examples to the test suite in the future (i.e.: just check they run okay, at least with no crashes).

Peque:
Fixes #162.

docs/source/introduction.rst
```diff
@@ -44,6 +44,11 @@ Running this example from your terminal should simply show you a log message
 saying `Hello world!` but, what exactly is happening there?
 
 
+.. note:: The ``ns.shutdown()`` call is present because we don't want examples
+   to run indefinetely. For more information on that method, refer to the
+   :ref:`shutdown_nameserver` section.
```
Peque: I would rather point to the current "Shuttting down" section in the "Considerations" (as shutting-down is not specific to distributed systems).

docs/source/distributed_systems.rst
```diff
@@ -59,6 +59,26 @@ address by its alias:
...
+
+We can finish the execution of a name server through the
+:func:`nameserver.shutdown <osbrain.nameserver.shutdown>` method.
+
+Shutting down a name server will also shut down any registered agents on it.
+This is specially useful when within our system, we are going to have
+various independent groups of agents. That way, we can register all agents of
+each group in its own name server and easily shut down each group by a
+:func:`nameserver.shutdown <osbrain.nameserver.shutdown>` call to its
+name server.
```
Peque: I would rather merge this with the current "Shutting down" section in "Considerations". I think having it there makes more sense as shutting down is not specific to distributed systems.

docs/source/distributed_systems.rst
```diff
@@ -59,6 +59,26 @@ address by its alias:
...
+
+Note that even if a name server is running without any registered agents,
+it will be kept alive until a call to
+:func:`nameserver.shutdown <osbrain.nameserver.shutdown>`.
```
Peque: This last paragraph seems superfluous to me, but if you think it could help someone, keep it there.

Flood1993:
No, examples didn't run OK in Windows before those changes neither. I fixed the pep8 issue and added one example regarding multiple name servers.

docs/source/considerations.rst
```diff
@@ -93,17 +93,38 @@ In fact, the only difference is the binding from Alice, in which we are using
...
+Shutting down a name server will result in all agents registered in that name
+server being shut down as well. This allows us to easily shutdown groups of
+agents at the same time, by registering them in different name servers when
+created.
+
+.. literalinclude:: ../../examples/various_nameservers.py
+
+.. note:: As seen in the example above, that two agents are registered on
+   different name servers does not mean that we cannot stablish connections
+   between them.
```
Peque: Hmmmmm. I don't like this section very much...

I feel it is too complicated. Maybe the first paragraph ("Sometimes, we will want...") and the third paragraph ("We learned earlier...") could be completely removed, as they do not add useful information (in my opinion we should try to keep documentation clean and short). If we remove those paragraphs, then maybe we can remove the subsections too, and just keep a single "Shutting down" title/section.

The first example, which is shorter and simpler already shows the use of a single agent shutdown and the name server shutdown. Is it really necessary to add a more complex example just after? Couldn't we just note in the documentation how the `ns.shutdown()` call will shutdown all remaining registered agents in the name server? (just like before these changes)

I feel the second example is too complicated (timers, PUB-SUB, topics...) just to show how to use the shutdown methods. Maybe it is clear enough with the first example.

I wouldn't say no to adding a note (like the one you have there) in which we mention that "note how you can use multiple name servers to register different groups of agents in order to shutdown the whole group more easily. Of course, you do not need to have two agents registered in the same name server in order to create connections between them.". Or something like that. But I see that more like a note, just to make it clear, not something we should explicitly show with an example (I would assume the user understands how to do that if they want to).

What do you think?

docs/source/introduction.rst
```diff
@@ -44,6 +44,11 @@ Running this example from your terminal should simply show you a log message
 saying `Hello world!` but, what exactly is happening there?
 
 
+.. note:: The ``ns.shutdown()`` call is present because we don't want examples
+   to run indefinetely. For more information on that method, refer to the
+   :ref:`shutting_down` section.
```
Peque: Maybe "is here" rather than "is present"? Sounds weird to me as it is right now, but maybe it is correct. If "is present" sounds better to you, just keep it that way. :blush:

examples/dispatcher.py
```diff
@@ -33,3 +34,7 @@ def request_work(agent):
         worker.connect(results_addr, alias='results')
         worker.connect(dispatcher_addr, alias='dispatcher')
         worker.each(0., request_work)
+
+    time.sleep(20)
+
+    ns.shutdown()
```
Peque: Maybe:

```python
while len(ns.agents()) > 2:                                                     
    time.sleep(0.1)                                                               
                                                                                    
ns.shutdown()
```

To shutdown as soon as possible (and notice the difference between this example and the ventilator example.

examples/ventilator.py
```diff
@@ -26,3 +26,7 @@ def heavy_processing(agent, message):
 
     for task in [5, 1, 1, 1, 1, 5, 1, 1, 1, 1]:
         ventilator.send('push', task)
+
+    time.sleep(12)
+
+    ns.shutdown()
```
Peque: Maybe here I would suggest you the same thing as with the dispatcher example.

examples/various_nameservers.py
```diff
@@ -0,0 +1,42 @@
...
+    annoyer_grape = run_agent('annoyer1', nsaddr=ns_annoy.addr())
+    annoyer_grape.connect(addr, alias='annoy')
+    annoyer_grape.each(0.2, annoy, 'grape')
+
+    time.sleep(1)
+    # Shutting down the annoying agents through their name server
+    ns_annoy.shutdown()
+
+    time.sleep(1)
+    ns.shutdown()
```
Peque: As I told you, I am not sure this example is good. Maybe it is too complicated for what it tries to show. And maybe there is no need to create an example for this?

docs/source/considerations.rst
Flood1993: Yes, now that you say it, it is true that the second example is too complicated. I will remove it from the documentation and make it simpler as you suggest. However, I'm thinking about keeping it on the ``examples/`` folder, since it wouldn't hurt having it there. What do you think?

Peque: Definitely, the examples directory can hold many examples, even complex cases. Only not all need to be shown (maybe just linked) from the documentation. Dispatcher and ventilator are not displayed in the documentation for example (I think). Anyway, if the purpose of that example is to show the shutdown of a group of agents, try to constrain the example (i.e.: avoid, if possible, timers, pub-sub, topics...). :wink:

Flood1993: Okay, I will try making the example simpler 👍

Flood1993: Mmmmhh... I can't think of a better example than the one I currently have. However, I think it could be useful (specially for "learning by example"). I've added some docstrings stating that the example is "relatively complex". However, if you feel the example and some explanations of what we are doing on it. In any case, if you feel the example should not be added to the project I've no problem taking it out.

PR #184, opened 2017-07-27
Document name server proxy creation

### LINKED ISSUE #169: Document name server proxy creation in distributed systems
Maybe this should be the first section in the "Distributed systems" documentation. Why is it good to have a unique name server?

```python
ns = NSProxy(nsaddr='127.0.0.1:1234')
```

It could also be documented earlier, in considerations, as this can also be used in non-distributed systems, where we simply use multiple scripts to configure our system (that, however, should maybe be discouraged in a note, as you can always do the same from one script).

Peque:
Fixes #169.

### DIFF
--- a/docs/source/considerations.rst
+++ b/docs/source/considerations.rst
@@ -158,3 +158,36 @@ running the agent for exactly that:
 As you can see, this parameter accepts a dictionary in which the keys are the
 name of the attributes to be set in the agent and the values are the actual
 values that this attributes will take.
+
+
+.. _creating_proxies_to_existing_name_servers:
+
+Creating proxies to existing name servers
+=========================================
+
+Many times, specially if we are not working with distributed systems, we want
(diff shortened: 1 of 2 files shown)


Peque:
Coverage failure not really related to this PR... https://github.com/opensistemas-hub/osbrain/issues/185

docs/source/considerations.rst
```diff
@@ -158,3 +158,36 @@ running the agent for exactly that:
...
+.. code-block:: python
+
+   from osbrain import NSProxy
+
+
+   ns = NSProxy(nsaddr='127.0.0.1:1234')
+
+Note how we need to specify the name server address.
+
+.. note:: This might be usefull for attaching yourself to an already-running
```
Flood1993: useful

Peque: Thanks! :blush:

PR #189, opened 2017-08-04
Updating documentation a bit

### LINKED ISSUE #168: Document `set_attr`
Currently there is no reference to this function. Users might be tempted to execute from a proxy:

```python
agent_proxy.attribute = value
```

Explain why this is wrong and how to use `set_addr` instead.

Refer to the OOP section in the documentation for more complex cases. Consider extending this documentation as well: how to use handlers when they are methods of the agent, how to retrieve an address by its alias...

Peque:
Fixes #168.

### DIFF
--- a/docs/source/considerations.rst
+++ b/docs/source/considerations.rst
@@ -136,11 +136,12 @@ however you may notice some differences:
    #. ``on_init()``: which is executed on initialization and will, in this
       case, simply bind a ``'PUSH'`` communication channel.
    #. ``hello()``: which simply logs a *Hello* message when it is executed.
-
 #. When connecting *Bob* to *Alice*, we need the address where *Alice* binded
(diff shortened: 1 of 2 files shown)


docs/source/considerations.rst
```diff
@@ -111,6 +111,8 @@ agents at the same time.
    name servers.
 
 
+.. _oop:
```
Flood1993: Is this necessary? I think we are not indexing to that label (even though I'm not exactly sure what do those two dots do)

Peque: Actually I am referring to that label in `...then you might use `oop` instead...`.

Flood1993: Oh, yes, I was searching with the underscore 😅

PR #198, opened 2017-09-05
Re-integrate pytest-timeout plugin

Peque:
There used to be a global timeout for tests but the plugin did not get along well with `xdist`, so I decided to go with parallel tests rather than timed-out tests.

Maybe we should reconsider this or at least check if the situation has changed.

References:

- https://bitbucket.org/pytest-dev/pytest-timeout/issues/8/internalerror-valueerror-signal-only-works
- https://github.com/pytest-dev/pytest-xdist/issues/117
- https://github.com/pytest-dev/pytest-xdist/issues/220
- https://github.com/pytest-dev/pytest-xdist/issues/133
- https://github.com/pytest-dev/pytest-xdist/issues/60


Peque:
See this as well: https://github.com/pytest-dev/execnet/issues/43

Issue #182, opened 2017-07-26
Travis builds ocasionally stalling


Flood1993:
This problem could be not ours. Just in case, reference here any stalled builds. If the build is to be relaunched, upload the raw log file instead of linking the URL.

Add the new ones at the end.

https://travis-ci.org/opensistemas-hub/osbrain/jobs/257634628
[travis_stall_01.txt](https://github.com/opensistemas-hub/osbrain/files/1176201/travis_stall_01.txt)
[travis_stall_02.txt](https://github.com/opensistemas-hub/osbrain/files/1190528/travis_stall_02.txt)
[log.txt](https://github.com/opensistemas-hub/osbrain/files/1199970/log.txt)
[log2.txt](https://github.com/opensistemas-hub/osbrain/files/1199971/log2.txt)
[log3.txt](https://github.com/opensistemas-hub/osbrain/files/1199999/log3.txt)
[log4.txt](https://github.com/opensistemas-hub/osbrain/files/1200147/log4.txt)
[travis_stall_03.txt](https://github.com/opensistemas-hub/osbrain/files/1276916/travis_stall_03.txt)
[travis_stall_04.txt](https://github.com/opensistemas-hub/osbrain/files/1276962/travis_stall_04.txt)
https://travis-ci.org/opensistemas-hub/osbrain/jobs/277983732
https://travis-ci.org/opensistemas-hub/osbrain/jobs/283438775
https://travis-ci.org/opensistemas-hub/osbrain/jobs/286774257
https://travis-ci.org/Peque/osbrain/jobs/289552073
[log5.txt](https://github.com/opensistemas-hub/osbrain/files/1425196/log5.txt)
[log6.txt](https://github.com/opensistemas-hub/osbrain/files/1425197/log6.txt)

Peque:
Moving this to 0.5.0 to see if we can solve it before next release.

Peque:
Let us keep an eye on this to see if it stalls again after fixing #190.

Flood1993:
Seems like I got another stall: https://travis-ci.org/Flood1993/osbrain/builds/271982282?utm_source=email&utm_medium=notification

Peque:
I guess we should verify all tests to see if there are possible infinite loops (there probably are).

We could also have a look at #189.

Peque:
Maybe this is related: https://github.com/pytest-dev/execnet/issues/43

Peque:
[...]

PR #207, opened 2017-10-31
Make shutdown process a bit more reliable

Peque:
Maybe fixes_ #182.

For now 25/27 runs passed (considering a run to be one of the three jobs Travis executes for each Python version).

The 2 only errors are due to `test_nameserver_proxy_shutdown_with_many_agents()` failing. Will investigate that failure separately. No other failures found.

I still consider this to be far more reliable than the current process. Also, the error is now at least displayed in the Travis log (before we used to get a stall but no errors, which is much worse).

### DIFF
--- a/osbrain/nameserver.py
+++ b/osbrain/nameserver.py
@@ -9,6 +9,7 @@ import multiprocessing
 
 import Pyro4
 from Pyro4.naming import BroadcastServer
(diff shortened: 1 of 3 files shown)


Peque:
@Flood1993 I think this is a step forward. :blush:

Issue #210, opened 2017-11-10
Crash on timer tests


Flood1993:
Seems like ``test_timer_each`` is still occasionally crashing on teardown.

- https://travis-ci.org/opensistemas-hub/osbrain/jobs/300138642
- https://travis-ci.org/Peque/osbrain/jobs/300138637

PR #211, opened 2017-11-10
Test safe calls on non-running agents

Peque:
While reviewing #158 found a piece of code in `agent.py` that was not covered and fixing it was very simple...


Flood1993:
Test not fully covered 😄 . Maybe add a call to ``my_name()`` before raising the exception?

Peque:
@Flood1993 Yeah, already fixed, I think. Travis will have the last word. :-P

Peque:
"very simple" :joy:

Peque:
Seems like one test failed with: https://travis-ci.org/opensistemas-hub/osbrain/jobs/300138642

Updated #210 as it is probably related.

PR #217, opened 2018-01-15
Documentation spelling and grammar

ocaballeror:
I went through all the docs and found this series of typos and grammar mistakes.

PR #215, opened 2017-12-15
Fixtests

Peque:
We can discuss this via Hangouts next Monday/Tuesday if you can. Otherwise in 2018. :blush:

Hopefully:

- Fixes #210 
- Fixes #212

We will need to execute the tests a couple of times to make sure. I already did that in my fork and so far it looks good.

PR #218, opened 2018-01-16
Spelling and grammar revision

ocaballeror:
Reopening pull request #217 from a different branch.

I went through all the docs and found this series of typos and grammar mistakes.

### DIFF
--- a/docs/source/distributed_systems.rst
+++ b/docs/source/distributed_systems.rst
@@ -26,19 +26,18 @@ Binding with distributed systems
 ================================
 
 Working with distributed systems is easy, you just basically need to make
-sure you use TCP transport and to specify the network interface where you
-want to bind.
+sure you use TCP transport and to specify the network interface IP address
+that you want to bind to.
 
 This can be done when binding (note that we are using `127.0.0.1` but you
-may use any other interface available in your system):
+may use any other IP address available in your system):
 
 .. code-block:: python
 
    address = agent.bind('PULL', transport='tcp', addr='127.0.0.1')
 
-Defining just the network interface means the agent will bind to a random
-port, which is usually fine. However, you can also specify the port to bind
-to:
+Defining just the IP address means the agent will bind to a random port, which
+is usually fine. However, you can also specify the port to bind to:
 
 .. code-block:: python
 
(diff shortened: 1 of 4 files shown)


Peque:
Don't worry for the Travis fail, for that we have #215.

docs/source/distributed_systems.rst
```diff
@@ -26,19 +26,18 @@ Binding with distributed systems
...
 
 Working with distributed systems is easy, you just basically need to make
-sure you use TCP transport and to specify the network interface where you
-want to bind.
+sure you use TCP transport and to specify the IP address that you want to bind
+to.
 
 This can be done when binding (note that we are using `127.0.0.1` but you
-may use any other interface available in your system):
+may use any other address available in your system):
```
Peque: Agree with this change: address is better than interface. However, maybe it is good to point out that the IP address cannot be any IP address, but only those corresponding to your network interfaces.

What do you think about using "network interface IP address" instead. Just for completeness. :blush:

docs/source/distributed_systems.rst
```diff
@@ -77,22 +76,23 @@ Proxies and multiple scripts
...
+  those as well.
 - From the second script, you access agents that you created on the first
-  and except that they are all ready and completely configured. But it turns
+  and expect them to be ready and completely configured. But it turns
   out that might not be the case.
 
-There is a very simple way to avoid this issue. Do create the agents from
-different scripts, but try to do all the configuration just from one of them.
+There is a very simple way to avoid this issue. You may create the agents from
+different scripts, but try to do all the configuration from one of them.
```
Peque: Although this was not clear before either, maybe we could say "do all the configuration from one of the scripts" instead of "from one of them", just to make sure the user does not interpret "them" as the agents instead of the scripts.

docs/source/distributed_systems.rst
```diff
@@ -77,22 +76,23 @@ Proxies and multiple scripts
...
 Remember that proxies allow you to treat remote agents just like local objects,
-which means you do not really mind if they are running in one machine or
+which means it doesn't really matter if they are running in one machine or
```
Peque: We try to avoid contractions in the documentation. So I would rather write "does not really matter". :wink:

examples/timer_each.py
```diff
@@ -21,7 +21,7 @@ def annoy(agent, say, more=None):
     apple.connect(addr, handler=log_message)
 
     # Multiple timers with parameters
-    orange.each(1., annoy, 'Hey')
+    orange.each(1.0, annoy, 'Hey')
```
Peque: This is unnecessary, but if it annoys you...

Be prepared to find a couple of them in the code though. :stuck_out_tongue_winking_eye:

ocaballeror: I changed it thinking it was an error. Even though, the syntax is valid, this still looks better in my opinion. I can live with it, though

docs/source/distributed_systems.rst
ocaballeror: I'll keep that in mind.
Still, I think "does not really matter" sounds better than "you do not really mind"

Peque: Sure! Go ahead with the change. Just avoid the contraction. :blush:

examples/timer_each.py
Peque: Leave it as `1.0` if you want. Might be clearer for newcomers as well. :+1:

docs/source/distributed_systems.rst
ocaballeror: "create the agents from different scripts, but try to do all the configuration from one of the scripts" still sounds somewhat weird to me. What if we rephrased the whole thing to something like:

"To avoid this issue, try to do all the configuration of agents from a single script, even if those agents are created from different ones". Does that sound better?

docs/source/distributed_systems.rst
ocaballeror: Sounds good to me.

docs/source/distributed_systems.rst
Peque: Much better. :blush:

docs/source/distributed_systems.rst
```diff
@@ -26,19 +26,18 @@ Binding with distributed systems
...
 
 Working with distributed systems is easy, you just basically need to make
-sure you use TCP transport and to specify the network interface where you
-want to bind.
+sure you use TCP transport and to specify the IP address that you want to bind
+to.
 
 This can be done when binding (note that we are using `127.0.0.1` but you
-may use any other interface available in your system):
+may use any other network interface IP address available in your system):
```
Peque: You put "network interface IP address" here, but not in the paragraph above. I would rather put that in both or, at least, in the first paragraph. :wink:

PR #221, opened 2018-01-17
Add a parameter to run the agent on proxy creation

### LINKED ISSUE #153: Proxy method to wait for agent to be running
Either create this method (does not break anything) or change the default behavior when creating proxies to agents so that these proxies wait for the agent not only to be ready but running.

What is the best option?

Update
---------

The issue is with proxy creation, that means, when creating a `osbrain.Proxy(...)` object. It currently waits for the agent to be ready (i.e.: registered in the name server and with the Pyro4 server running). However, it does not wait for it to be running, which should probably be the default behavior.

ocaballeror:
I'm not sure I understood issue #153 correctly, but this is what I came up with

**Edit**: fixes #153.

### DIFF
--- a/docs/source/distributed_systems.rst
+++ b/docs/source/distributed_systems.rst
@@ -96,3 +96,6 @@ script, even if those agents are created from different ones.
 Remember that proxies allow you to treat remote agents just like local objects,
 which means it does not really matter if they are running in one machine or
 another.
+
+.. note:: If you want to make sure the agent is running before using it, you
+   can use the `Proxy.wait_for_running()` method.
--- a/osbrain/proxy.py
+++ b/osbrain/proxy.py
@@ -100,6 +100,35 @@ class Proxy(Pyro4.core.Proxy):
             raise
         return True
 
+    def wait_for_running(self, timeout=3.):
+        """
+        Wait until the agent is running.
+
+        Parameters
+        ----------
+        timeout : float
+            Raise and exception if the agent is not running after this number
+            of seconds. Use a negative value to wait forever.
+
+        Raises
+        ------
+        TimeoutError
+            If the agent is not running after the given timeout.
+
+        Returns
+        -------
+        Proxy
+            The object itself.
+        """
+        time0 = time.time()
+        while not self.get_attr('running'):
+            if timeout >= 0 and time.time() - time0 > timeout:
+                msg = 'Timed out while waiting for the agent to be running'
+                raise TimeoutError(msg)
+            time.sleep(.01)
+
+        return self
+
     def __getstate__(self):
         return super().__getstate__() + \
             (self._next_oneway, self._default_safe, self._safe)
--- a/osbrain/tests/test_proxy.py
+++ b/osbrain/tests/test_proxy.py
@@ -5,10 +5,13 @@ import time
 
 import pickle
 import pytest
+from threading import Timer
+
 
 import osbrain
 from osbrain import run_agent
 from osbrain import Agent
+from osbrain import AgentProcess
 from osbrain import Proxy
 from osbrain.proxy import locate_ns
 from osbrain.helper import wait_agent_attr
@@ -117,6 +120,55 @@ def test_agent_proxy_initialization_timeout(nsproxy):
     run_agent('foo')
     with pytest.raises(TimeoutError):
         InitTimeoutProxy('foo', timeout=1.)
+
+
+@pytest.mark.parametrize('timeout', [-1, 1])
+def test_agent_proxy_wait_running(nsproxy, timeout):
+    """
+    Using `wait_for_running` on a proxy after initialization should block until
+    the agent is running or time out.
+    """
+    AgentProcess('agent').start()
+
+    # Get "offline" proxy
+    agent = Proxy('agent')
+    time0 = time.time()
+    Timer(abs(timeout) / 2, agent.run).start()
+
+    proxy = Proxy('agent').wait_for_running(timeout=timeout)
+    elapsed = time.time() - time0
+    assert proxy.ping() == 'pong'
+    assert elapsed >= abs(timeout) / 2
+    assert elapsed <= abs(timeout)
+
+
+def test_agent_proxy_wait_running_0_seconods(nsproxy):
+    """
+    Using `wait_for_running` on a proxy after initialization should block until
+    the agent is running or time out.
+    """
+    run_agent('agent')
+
+    proxy = Proxy('agent').wait_for_running(timeout=0)
+    assert proxy.ping() == 'pong'
+
+
+@pytest.mark.parametrize('timeout', [0, 1])
+def test_agent_proxy_wait_running_timeout(nsproxy, timeout):
+    """
+    Check that the `wait_for_running` method times out if the agent is not
+    running after the specified number of seconds.
+    """
+    AgentProcess('agent').start()
+
+    time0 = time.time()
+    with pytest.raises(TimeoutError) as error:
+        Proxy('agent').wait_for_running(timeout=timeout)
+    elapsed = time.time() - time0
+
+    assert "Timed out" in str(error.value)
+    assert elapsed >= timeout
+    assert elapsed < timeout + .5
 
 
 def test_agent_proxy_nameserver_address(nsproxy):
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -1789,6 +1789,6 @@ def run_agent(name, nsaddr=None, addr=None, base=Agent, serializer=None,
                  attributes=attributes).start()
     proxy = Proxy(name, nsaddr, safe=safe)
     proxy.run()
-    while not proxy.get_attr('running'):
-        time.sleep(0.01)
+    proxy.wait_for_running()
+
     return proxy


ocaballeror:
My bad, some tests are failing. I'll update the branch when everything works again.

Peque:
It seems Travis is down... :confused:

Yeah, I think you misunderstood the issue. When in doubt... just ask! :wink: It is only normal to misunderstand some issues/conversations if you were not there the day they started. It is also probably our fault for not being clearer in the description.

Updated the issue to, hopefully, make it more clear.

Also, all changes should have tests validating the new functionality, so have a look at the test suite and create a test that checks what you want to implement. You might want to create the tests first, let me review it, and then implement the fix that makes the new tests pass. :wink:

osbrain/proxy.py
```diff
@@ -100,6 +100,30 @@ def _ready_or_timeout(self, time0, timeout):
             raise
         return True
 
+    def wait_for_running(self, timeout=0):
```
Peque: Even though the type is well documented, I would rather use `0.` here. The default type can also help understanding which is the expected type for a parameter. :wink:

osbrain/proxy.py
```diff
@@ -287,6 +311,8 @@ def proxy(self, name, timeout=3.):
             Proxy name, as registered in the name server.
         timeout : float
             Timeout, in seconds, to wait until the agent is discovered.
+        running : bool
+            Whether to wait for the agent to be running or not.
```
Peque: Any particular reason you changed the default behavior in this method?

I mean, why do you think `Proxy()` should behave differently than `ns.proxy()` by default?

osbrain/proxy.py
Peque: Also, I think a default value that is not 0., but greater, makes more sense. Waiting really forever is rarely useful.

Maybe `3.`? I think we currently use that timeout by default in other operations.

osbrain/proxy.py
```diff
@@ -100,6 +100,30 @@ def _ready_or_timeout(self, time0, timeout):
...
 
+    def wait_for_running(self, timeout=0):
+        """
+        Wait until the agent is running.
+
+        Parameters
+        ----------
+        timeout : float
+            If different than 0, raise an error if the agent is not running
+            after that number of seconds.
```
Peque: What if we say: "if greater or equal than 0" (or "if not negative")?

0 seconds timeout could make sense as well. In example: just see if the agent is running once and otherwise immediately raise. -1 seconds timeout makes no sense and is many times used for non-negative parameters that could be ignored.

osbrain/proxy.py
```diff
@@ -277,7 +301,7 @@ def release(self):
         """
         self._pyroRelease()
 
-    def proxy(self, name, timeout=3.):
+    def proxy(self, name, timeout=3., running=True):
```
Peque: Instead of `running=True` I think we should have a more consistent API and put `timeout=X.`.

osbrain/tests/test_proxy.py
```diff
@@ -119,6 +122,42 @@ def ping(self):
         InitTimeoutProxy('foo', timeout=1.)
 
 
+def test_agent_proxy_wait_running(nsproxy):
+    """
+    A proxy initialized with running=True should wait for a number of seconds
+    until the agent is running
```
Peque: I think this docstring is outdated. :wink:

osbrain/tests/test_proxy.py
```diff
@@ -119,6 +122,42 @@ def ping(self):
...
+    ap = AgentProcess('agent', nsaddr=nsproxy.addr())
+    ap.start()
+
+    # Get "offline" proxy
+    agent = Proxy('agent', nsproxy.addr())
+    Timer(.5, agent.run).start()
+
+    # Wait long enough for the agent to start running
+    proxy = Proxy('agent', nsproxy.addr(), timeout=1).wait_for_running()
+    assert proxy.ping() == 'pong'
```
Peque: To simplify this test, I think you could remove the `nsproxy.addr()` from all calls.

osbrain/tests/test_proxy.py
```diff
@@ -119,6 +122,42 @@ def ping(self):
...
+
+
+def test_agent_proxy_wait_running_timeout(nsproxy):
+    """
+    A proxy initialized with running=True should wait for a number of seconds
+    until the agent is running
```
Peque: I think this docstring is outdated too.

osbrain/tests/test_proxy.py
```diff
@@ -119,6 +122,42 @@ def ping(self):
...
+
+    # Get "offline" proxy
+    agent = Proxy('agent', nsproxy.addr())
+
+    # Wait for too long before starting the agent
+    Timer(3, agent.run).start()
+    with pytest.raises(TimeoutError) as error:
+        Proxy('agent', nsproxy.addr()).wait_for_running(timeout=.5)
+
+    assert "Timed out" in str(error.value)
```
Peque: To simplify this test, I think you could remove the `nsproxy.addr()` from all calls.

osbrain/tests/test_proxy.py
```diff
@@ -119,6 +122,42 @@ def ping(self):
...
+    until the agent is running
+    """
+    ap = AgentProcess('agent', nsaddr=nsproxy.addr())
+    ap.start()
+
+    # Get "offline" proxy
+    agent = Proxy('agent', nsproxy.addr())
+
+    # Wait for too long before starting the agent
+    Timer(3, agent.run).start()
```
Peque: I think you could further simplify the test by simply not calling to `Timer().start()`. No need to start the agent.

osbrain/proxy.py
ocaballeror: I thought the timeout should be optional,  but now that you mention it, if the agent  is not ready after 3 seconds, you probably have some other problem, and the agent is likely to never be ready at all. I'll change it to 3, then :smile:

osbrain/proxy.py
ocaballeror: I didn't think 0 seconds made any sense, but OK, I'll change it to "less than zero"

osbrain/proxy.py
ocaballeror: Same as above,  I forgot to remove this part.

osbrain/proxy.py
ocaballeror: Leftover code. It doesn't really make sense with the new implementation.

osbrain/tests/test_proxy.py
ocaballeror: Definitely :smile:

osbrain/tests/test_proxy.py
ocaballeror: I'm not really sure when this parameter is necessary and when it isn't, so I always add it in for safety. How does the proxy initalization know which nameserver to connect to?

osbrain/tests/test_proxy.py
ocaballeror: You got me again

osbrain/tests/test_proxy.py
ocaballeror: You're right.  I just thought this would be easier to understand, but it makes sense to remove it

osbrain/proxy.py
```diff
@@ -100,6 +100,30 @@ def _ready_or_timeout(self, time0, timeout):
...
+        Parameters
+        ----------
+        timeout : float
+            If not negative, raise an error if the agent is not running
+            after that number of seconds.
+
+        Returns
+        -------
+        Proxy
+            The object itself
```
Peque: Just a dot after the sentence. :blush:

osbrain/tests/test_proxy.py
```diff
@@ -119,6 +122,41 @@ def ping(self):
         InitTimeoutProxy('foo', timeout=1.)
 
 
+def test_agent_proxy_wait_running(nsproxy):
+    """
+    Using wait_for_running on a proxy after initialization should block until
```
Peque: You could use backticks ( \` ) around \`wait_for_running\` in the docstring. You could also add a dot and/or parenthesis as in `.wait_for_running()`, but that only in case you prefer that style.

osbrain/tests/test_proxy.py
```diff
@@ -119,6 +122,41 @@ def ping(self):
...
+
+    # Wait long enough for the agent to start running
+    proxy = Proxy('agent', timeout=1).wait_for_running()
+    assert proxy.ping() == 'pong'
+
+
+def test_agent_proxy_wait_running_timeout(nsproxy):
+    """
+    Using wait_for_running on a proxy after initialization should block until
+    the agent is running or time out.
```
Peque: You could use backticks ( \` ) around \`wait_for_running\` in the docstring. You could also add a dot and/or parenthesis as in `.wait_for_running()`, but that only in case you prefer that style.

osbrain/tests/test_proxy.py
```diff
@@ -119,6 +122,41 @@ def ping(self):
...
+    """
+    ap = AgentProcess('agent')
+    ap.start()
+
+    # Get "offline" proxy
+    Proxy('agent', nsproxy.addr())
+
+    # Wait for too long before starting the agent
```
Peque: No need for this comment now. :wink:

osbrain/tests/test_proxy.py
```diff
@@ -119,6 +122,41 @@ def ping(self):
...
+    with pytest.raises(TimeoutError) as error:
+        Proxy('agent', nsproxy.addr()).wait_for_running(timeout=.5)
```
Peque: You  could test with `timeout=0.` too and see if it passes. :wink:

osbrain/tests/test_proxy.py
```diff
@@ -119,6 +122,41 @@ def ping(self):
...
+
+    assert "Timed out" in str(error.value)
```
Peque: Just if you want: you could assert the *approximate* time passed between the `Proxy()` call and the raised exception.

In example: for 0 seconds timeout check that less than 1 second passed. For 1 seconds timeout check that more than 1 seconds passed but less than 2 seconds.

Just to make the test a bit more complete. :blush:

osbrain/tests/test_proxy.py
Peque: They have in reST/Sphinx, which is the format we are using for docs generation. Things surrounded by backquotes can be interpreted as code/names and even automatically linked.

osbrain/proxy.py
```diff
@@ -100,6 +100,30 @@ def _ready_or_timeout(self, time0, timeout):
...
 
+    def wait_for_running(self, timeout=3.):
+        """
+        Wait until the agent is running.
+
+        Parameters
+        ----------
+        timeout : float
+            If not negative, raise an error if the agent is not running
+            after that number of seconds.
```
Peque: Maybe: "Raise an exception if the agent is not running after this number of seconds. Use a negative value to wait indefinitely." (o forever). What you think?

osbrain/proxy.py
```diff
@@ -100,6 +100,30 @@ def _ready_or_timeout(self, time0, timeout):
...
+        Returns
+        -------
+        Proxy
+            The object itself.
+        """
+        time0 = time.time()
+        while not self.get_attr('running'):
+            if timeout >= 0 and time.time() - time0 > timeout:
+                raise TimeoutError('Timed out while waiting for the agent to\
+                        run')
```
Peque: Are you sure this exception will be appropriately formatted? Try it to see it for yourself.

If the printed message is not correctly formatted I'd rather use something like:

```python
# Shorter message, as the "TimeoutError" already tells you it is a timeout.
raise TimeoutError('Agent is not running')
```

Or:

```python
# Line split differently
raise TimeoutError('Asdf foo '
                   'bar qwer')
```

Or even:

```python
# Variable for the error message (more useful if you were to format it with variables)
error = 'Error message'
raise TimeoutError(error)
```

osbrain/tests/test_proxy.py
```diff
@@ -119,6 +122,75 @@ def ping(self):
...
+def test_agent_proxy_wait_running_timeout_0_seconds(nsproxy):
+    """
+    Check that the `wait_for_running` times out when the specified number of
+    seconds is 0
+    """
+    ap = AgentProcess('agent')
+    ap.start()
+
+    # Get "offline" proxy
+    Proxy('agent', nsproxy.addr())
```
Peque: No need to specify the `nsproxy.addr()` here. :wink:

osbrain/tests/test_proxy.py
```diff
@@ -119,6 +122,75 @@ def ping(self):
...
+    # Get "offline" proxy
+    agent = Proxy('agent')
+    Timer(.5, agent.run).start()
+
+    # Wait long enough for the agent to start running
+    proxy = Proxy('agent', timeout=1).wait_for_running()
+    assert proxy.ping() == 'pong'
+
+
+def test_agent_proxy_wait_running_timeout(nsproxy):
```
Peque: This test and the last one share the same code. You should merge them into one that simply tests the TimeoutError exception when the agent is not running.

Use `@pytest.mark.parametrize` to change the timeout value (you can find a couple of examples in the osBrain test suite too).

ocaballeror: I didn't even know this was a possibility, but looking at the other tests, I can see it makes a lot of sense.

osbrain/proxy.py
ocaballeror: Sounds better. I'll change it

osbrain/proxy.py
Peque: @ocaballeror Could be redundant as we are already using `TimeoutError`, but is fine for me. :blush:

osbrain/tests/test_proxy.py
Peque: I guessed so. `pytest` is a great tool. :blush:

ocaballeror: Definitely. I'll have to investigate what other things can it do

osbrain/tests/test_proxy.py
```diff
@@ -119,6 +122,48 @@ def ping(self):
...
+    Timer(abs(timeout)/2, agent.run).start()
+
+    # Wait long enough for the agent to start running
+    proxy = Proxy('agent').wait_for_running(timeout=timeout)
+    assert proxy.ping() == 'pong'
+    if timeout != 0:
+        assert abs(timeout)/2 <= time.time() - time0 <= abs(timeout)
+
+
+# Can't use negatives here because they will never time out
```
Peque: I don't think this comment is necessary.

osbrain/tests/test_proxy.py
```diff
@@ -119,6 +122,48 @@ def ping(self):
...
+def test_agent_proxy_wait_running_timeout(nsproxy, timeout):
+    """
+    Check that the `wait_for_running` method times out if the agent is not
+    running after the specified number of seconds.
+    """
+    ap = AgentProcess('agent')
+    ap.start()
+
+    # Get "offline" proxy
+    Proxy('agent')
```
Peque: I think these two lines are not necessary/used.

osbrain/tests/test_proxy.py
```diff
@@ -119,6 +122,48 @@ def ping(self):
...
+
+# Can't use negatives here because they will never time out
+@pytest.mark.parametrize('timeout', [0., .5])
+def test_agent_proxy_wait_running_timeout(nsproxy, timeout):
+    """
+    Check that the `wait_for_running` method times out if the agent is not
+    running after the specified number of seconds.
+    """
+    ap = AgentProcess('agent')
+    ap.start()
```
Peque: You are not using `ap` variable, so maybe you can `AgentProcess('agent').start()` instead? (one line, no variable).

osbrain/tests/test_proxy.py
```diff
@@ -119,6 +122,48 @@ def ping(self):
...
+    # Get "offline" proxy
+    agent = Proxy('agent')
+    time0 = time.time()
+    Timer(abs(timeout)/2, agent.run).start()
+
+    # Wait long enough for the agent to start running
+    proxy = Proxy('agent').wait_for_running(timeout=timeout)
+    assert proxy.ping() == 'pong'
+    if timeout != 0:
+        assert abs(timeout)/2 <= time.time() - time0 <= abs(timeout)
```
Peque: I liked the `0.` timeout case in a separate case more, just like you had it before. I think the negative case should be in a separate case as well. Just maybe.

osbrain/tests/test_proxy.py
ocaballeror: True. I probably was using it before, and that's why it was left here.

osbrain/tests/test_proxy.py
ocaballeror: True as well. Probably leftover from the old implementation.

ocaballeror:
I reviewed the tests. The final distribution is:
* A parametrized test for  `wait_for_running` with 1 and -1
* A  separate one that tests 0 as timeout
* A third test that triggers a timeout, parametrized with 0 and 1. It didn't make sense to move 0 to a separate test here.

osbrain/tests/test_proxy.py
```diff
@@ -119,6 +122,51 @@ def ping(self):
...
+    AgentProcess('agent').start()
+
+    # Get "offline" proxy
+    agent = Proxy('agent')
+    time0 = time.time()
+    Timer(abs(timeout)/2, agent.run).start()
+
+    proxy = Proxy('agent').wait_for_running(timeout=timeout)
+    assert proxy.ping() == 'pong'
+    assert abs(timeout)/2 <= time.time() - time0 <= abs(timeout)
```
Peque: I would rather always keep spaces around operators (i.e.: around the `/` here and in line 136). But I see the PEP8 and Flakes tests not complaining so...

Anyway it could be better, for consistency with the rest of the code.

Peque: I'd rather separate the last assertion in two: one for the lower bound, one for the upper. Two checks ---> two assertions. I think it is better that way than adding multiple comparisons or operators.

osbrain/tests/test_proxy.py
```diff
@@ -119,6 +122,51 @@ def ping(self):
...
+    running after the specified number of seconds.
+    """
+    AgentProcess('agent').start()
+
+    time0 = time.time()
+    with pytest.raises(TimeoutError) as error:
+        Proxy('agent').wait_for_running(timeout=timeout)
+
+    assert "Timed out" in str(error.value)
+    assert time.time() - time0 >= timeout
```
Peque: It would be great if you could add another assertion for the upper bound. Right now the TimeoutError could be raised after 3 seconds for both tests, for example. You are just checking the assertion occurs after the timeout, but not when.

Maybe something like: `assert time.time() - time0 < timeout + .5`.

ocaballeror: Not very critical, but OK :ok_hand:

docs/source/distributed_systems.rst
```diff
@@ -96,3 +96,6 @@ script, even if those agents are created from different ones.
 Remember that proxies allow you to treat remote agents just like local objects,
 which means it does not really matter if they are running in one machine or
 another.
+
+.. note:: If you want to make sure the agent is running before using it, you
+  can use the `Proxy.wait_for_running()` method
```
Peque: Normally we use 3 spaces here (i.e.: second line aligned with "note::"). Just as recommended by Sphinx. This style is not checked, so feel free to leave it as-is if you want.

osbrain/proxy.py
```diff
@@ -100,6 +100,30 @@ def _ready_or_timeout(self, time0, timeout):
...
 
+    def wait_for_running(self, timeout=3.):
+        """
+        Wait until the agent is running.
+
+        Parameters
+        ----------
+        timeout : float
+            Raise and exception if the agent is not running after this number
+            of seconds. Use a negative value to wait forever.
```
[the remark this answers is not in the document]
ocaballeror: Makes sense. However, I still think the parameter should be documented. I'll see how I can add this without being redundant.

osbrain/tests/test_proxy.py
```diff
@@ -119,6 +122,55 @@ def ping(self):
...
+    # Get "offline" proxy
+    agent = Proxy('agent')
+    time0 = time.time()
+    Timer(abs(timeout) / 2, agent.run).start()
+
+    proxy = Proxy('agent').wait_for_running(timeout=timeout)
+    elapsed = time.time() - time0
+    assert proxy.ping() == 'pong'
+    assert abs(timeout) / 2 <= elapsed
+    assert elapsed <= abs(timeout)
```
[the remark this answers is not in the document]
ocaballeror: Looks nicer :smile:

PR #222, opened 2018-01-19
Don't allow multiple agents to have the same name

### LINKED ISSUE #219: It is possible to create multiple agents with the same name
Users shouldn't be doing this, but this behavior shouldn't be allowed in the first place. A RuntimeError is raised when multiple agents are created with the same name, but I suggest we throw an error ourselves right when the second agent is registered with an existing name.

### Example:
```python
from osbrain import run_agent

ns = run_nameserver()
run_agent('agent')
run_agent('agent')
run_agent('agent')

agent = ns.proxy('agent')
agent.log_info('Hello world!')

ns.shutdown()
```

ocaballeror:
Should fix #219


osbrain/tests/test_agent.py
```diff
@@ -127,6 +127,17 @@ def test_run_agent_initial_attributes_exception(nsproxy):
     assert 'Agent already has "name" attribute!' in str(error.value)
 
 
+def test_run_agents_same_name(nsproxy):
```
Peque: Maybe we could move this test to `test_nameserver.py`?

osbrain/tests/test_agent.py
```diff
@@ -127,6 +127,17 @@ def test_run_agent_initial_attributes_exception(nsproxy):
...
+    """
+    Check that the user cannot create two agents with the same name. A
+    RuntimeError should be raised.
+    """
+    with pytest.raises(RuntimeError) as error:
+        run_agent('name')
+        run_agent('name')
+    assert('name already registered') in str(error.value)
```
Peque: Rather:

```python
assert 'name already registered' in str(error.value)
```

osbrain/tests/test_agent.py
```diff
@@ -127,6 +127,17 @@ def test_run_agent_initial_attributes_exception(nsproxy):
     assert 'Agent already has "name" attribute!' in str(error.value)
 
 
+def test_run_agents_same_name(nsproxy):
+    """
+    Check that the user cannot create two agents with the same name. A
+    RuntimeError should be raised.
+    """
+    with pytest.raises(RuntimeError) as error:
+        run_agent('name')
```
Peque: Get the first `run_agent()` call out of the `with` scope.

osbrain/agent.py
```diff
@@ -1697,10 +1697,15 @@ def run(self):
             self.queue.put(format_exception())
             return
 
+        uri = self._daemon.register(self.agent)
+        try:
+            ns.register(self.name, uri, safe=True)
+        except Pyro4.errors.NamingError as error:
+            self.queue.put(str(error))
+            return
```
Peque: Looks good. Two comments though:

- Maybe instead of `str(error)` you could use `format_exception()`?
- You could add a `finally: ns.release()` clause (and remove the current `ns.release()` call bellow).

ocaballeror:
Will do.
Please kill me if I push to upstream before origin again.

Peque:
@ocaballeror Remember to push everything here before merging (even rebases without squash). :wink:

PR #224, opened 2018-01-25
Add test to check we do not leave IPC sockets open

### LINKED ISSUE #216: Add test to check we do not leave IPC sockets open
Shutting down or killing an agent should always result in all sockets being closed (and their corresponding files being deleted).

There is no test currently for that, but we should create one. I think it can be reproduced if we remove the `close_all()` call in shutdown/kill.

In example: bind a socket to an unique address. Check the socket file is created. Crash the agent, shutdown the agent, kill the agent, block the agent... Shutdown the name server, shutdown the agent, kill the agent... All cases should result in the socket begin closed and the socket file being deleted.

ocaballeror:
In response to #216 

**Edit**: fixes #216.

### DIFF
--- a/osbrain/tests/test_agent_ipc_sockets.py
+++ b/osbrain/tests/test_agent_ipc_sockets.py
@@ -1,0 +1,83 @@
+"""
+Test file for everything IPC socket related
+"""
+import pytest
+
+from osbrain import run_agent
+from osbrain import run_nameserver
+from osbrain.helper import agent_dies
+
+from common import nsproxy  # pragma: no flakes
+
+
+def test_agent_close_ipc_socket_agent_shutdown(nsproxy):
+    """
+    Check that the socket is closed and the socket file removed after the agent
+    is shut down.
+    """
+    agent = run_agent('name')
+    address = agent.bind('PUSH')
+    agent.shutdown()
+
+    assert agent_dies('name', nsproxy, timeout=3)
+    assert not address.address.exists()
+
+
+def test_agent_close_ipc_socket_agent_kill(nsproxy):
+    """
+    Check that the socket is closed and the socket file removed after the agent
+    is killed.
+    """
+    agent = run_agent('name')
+    address = agent.bind('PUSH')
+    agent.kill()
+
+    assert agent_dies('name', nsproxy, timeout=3)
+    assert not address.address.exists()
+
+
+def test_agent_close_ipc_socket_agent_blocked_nameserver_shutdown():
+    """
+    Check that the socket is closed and the socket file removed when the name
+    server is shut down having a blocked agent.
+    """
+    def block(agent):
+        agent.send('out', 'blocking...')
+
+    ns = run_nameserver()
+    blocker = run_agent('blocker')
+    blocker.set_method(block)
+    addr = blocker.bind('PUSH', alias='out')
+    blocker.after(0, 'block')
+    ns.shutdown(timeout=1.)
+
+    assert not addr.address.exists()
+
+
+def test_agent_close_ipc_socket_agent_crash_nameserver_shutdown():
+    """
+    Check that the socket is closed and the socket file removed after the agent
+    crashes and the name server calls for shutdown.
+    """
+    ns = run_nameserver()
+    agent = run_agent('agent')
+    addr = agent.bind('PUSH', 'main')
+    with pytest.raises(RuntimeError):
+        agent.raise_exception()
+
+    ns.shutdown()
+
+    assert not addr.address.exists()
+
+
+def test_agent_close_ipc_socket_nameserver_shutdown():
+    """
+    Check that the socket is closed and the socket file removed after the name
+    server is shut down.
+    """
+    ns = run_nameserver()
+    agent = run_agent('agent')
+    addr = agent.bind('PUSH', 'main')
+    ns.shutdown()
+
+    assert not addr.address.exists()


osbrain/tests/test_agent.py
```diff
@@ -105,6 +105,88 @@ def test_agent_shutdown(nsproxy):
...
+    """
+    Check that the socket is closed and the socket file removed after the agent
+    crashes
+    """
+    agent = run_agent('agent')
+    addr = agent.bind('PUSH', 'main')
+    with pytest.raises(RuntimeError):
+        agent.raise_exception()
+
+    assert not addr.address.exists()
```
Peque: I am not sure we should be closing all sockets after an agent crashes. This prevents the user from being able to recover the agent very easily (i.e.: just re-running the main thread).

The idea I had in mind was rather to call a `nsproxy.shutdown()` after the crash (and/or an `agent.kill()`) and that should clean the sockets just fine.

But maybe we should close the sockets when a crash occurs straight away...

osbrain/tests/test_agent.py
```diff
@@ -105,6 +105,88 @@ def test_agent_shutdown(nsproxy):
     assert 'a0' not in nsproxy.list()
```
Peque: I would maybe rename all tests as `test_close_ipc_socket_xxxx` (adding the `ipc`).

Maybe we could even create a separate file for IPC tests (that would simplify the test suite Windows compatibility too).

osbrain/tests/test_agent.py
```diff
@@ -105,6 +105,88 @@ def test_agent_shutdown(nsproxy):
...
+
+    addr = alice.bind('REP', alias='main', handler=log)
+    bob.connect(addr, alias='main')
+
+    Timer(1, nsproxy.shutdown).start()
+    bob.send('main', 'Hello world')
+    rep = bob.recv('main')
+    print(rep)
+
+    assert not addr.address.exists()
```
Peque: I don't think this test is correct. If you are able to `print(rep)` it means the agent is not blocked. The idea here would be to make sure the agent is blocked when we call for `nsproxy.shutdown()` (and that the shutdown succeeds and the socket file gets closed).

Something like:

```python
def test_close_socket_agent_blocked(nsproxy):                                   
    """                                                                         
    Check that the socket is closed and the socket file removed when a blocked  
    agent is requested to shut down.                                            
    """                                                                         
    def block(agent):                                                           
        agent.send('out', 'blocking...')                                        
                                                                                
    blocker = run_agent('blocker')                                              
    blocker.set_method(block)                                                   
    addr = blocker.bind('PUSH', alias='out')                                    
    blocker.after(0, 'block')                                                   
    nsproxy.shutdown(timeout=1.)                                                
                                                                                
    assert not addr.address.exists()
```

osbrain/tests/test_agent.py
```diff
@@ -105,6 +105,88 @@ def test_agent_shutdown(nsproxy):
     assert 'a0' not in nsproxy.list()
 
 
+def test_close_socket_agent_shutdown(nsproxy):
+    """
+    Check that the socket is closed and the socket file removed after the agent
+    is shut down
```
Peque: Missing dot.

osbrain/tests/test_agent.py
```diff
@@ -105,6 +105,88 @@ def test_agent_shutdown(nsproxy):
...
+    agent = run_agent('name')
+    address = agent.bind('PUSH')
+    agent.shutdown()
+    assert not address.address.exists()
+
+
+def test_close_socket_agent_kill(nsproxy):
+    """
+    Check that the socket is closed and the socket file removed after the agent
+    is killed
```
Peque: Missing dot.

osbrain/tests/test_agent.py
```diff
@@ -105,6 +105,88 @@ def test_agent_shutdown(nsproxy):
...
+    agent = run_agent('name')
+    address = agent.bind('PUSH')
+    agent.kill()
+    assert not address.address.exists()
+
+
+def test_close_socket_agent_blocked(nsproxy):
+    """
+    Check that the socket is closed and the socket file removed when a blocked
+    agent is requested to shut down
```
Peque: Missing dot.

osbrain/tests/test_agent.py
```diff
@@ -105,6 +105,88 @@ def test_agent_shutdown(nsproxy):
...
+    rep = bob.recv('main')
+    print(rep)
+
+    assert not addr.address.exists()
+
+
+def test_close_socket_agent_crash(nsproxy):
+    """
+    Check that the socket is closed and the socket file removed after the agent
+    crashes
```
Peque: Missing dot.

osbrain/tests/test_agent.py
```diff
@@ -105,6 +105,88 @@ def test_agent_shutdown(nsproxy):
...
+    with pytest.raises(RuntimeError):
+        agent.raise_exception()
+
+    assert not addr.address.exists()
+
+
+def test_close_socket_nameserver_shutdown(nsproxy):
+    """
+    Check that the socket is closed and the socket file removed after the name
+    server is shut down
```
Peque: Missing dot.

osbrain/tests/test_agent.py
```diff
@@ -105,6 +105,88 @@ def test_agent_shutdown(nsproxy):
...
+    agent = run_agent('agent')
+    addr = agent.bind('PUSH', 'main')
+    nsproxy.shutdown()
+    assert not addr.address.exists()
+
+
+def test_close_socket_nameserver_crash(nsproxy):
+    """
+    Check that the socket is closed and the socket file removed after the name
+    server crashes
```
Peque: Missing dot.

osbrain/tests/test_agent.py
```diff
@@ -105,6 +105,88 @@ def test_agent_shutdown(nsproxy):
...
+    """
+    agent = run_agent('agent')
+    addr = agent.bind('PUSH', 'main')
+    with pytest.raises(RuntimeError):
+        agent.raise_exception()
+
+    assert not addr.address.exists()
+
+
+def test_close_socket_nameserver_shutdown(nsproxy):
```
Peque: Maybe all tests should start with `test_close_agent_socket_xxx` (even if `xxx` is `nameserver_shutdown`).

osbrain/tests/test_agent.py
```diff
@@ -105,6 +105,88 @@ def test_agent_shutdown(nsproxy):
...
+    """
+    Check that the socket is closed and the socket file removed after the name
+    server crashes
+    """
+    agent = run_agent('agent')
+    addr = agent.bind('PUSH', 'main')
+    with pytest.raises(RuntimeError):
+        nsproxy.raise_exception()
+
+    assert not addr.address.exists()
```
Peque: Maybe we will end-up separating this nameserver-crash test with a new issue. Not sure how we should handle this case. Maybe if the name server crashes we should shutdown everything.

osbrain/tests/test_agent.py
ocaballeror: Yes, I was thinking about putting them in a separate file, but it felt too drastic in the end, so I kept them in this one 😄 
I'll move them in the next fixup.

osbrain/tests/test_agent.py
ocaballeror: I really need  to look out for those dots 😄

osbrain/tests/test_agent.py
ocaballeror: Then there's something I don't understand about how the library works. I was thinking that if an agent was waiting to receive some message (therefore "blocked"), but before it came, a shutdown was issued, the agent would stop waiting ("unblock") and the execution would continue. Is there anything  wrong about my assumption?

Peque: Well, as I said, if you are able to print the received message, then that means the reception was not interrupted by the shutdown. :wink:

In this case what is probably happening is that the shutdown has a timeout and, therefore, the name server might be waiting a couple of seconds for the agent to shutdown before actually interrupting it.

In the example I showed you (which is also a bit shorter), we are sure the agent will never unblock naturally (the block will last forever). So the only way to shut everything down is actually interrupting the agent.

osbrain/tests/test_agent.py
ocaballeror: Are we assuming agents crash frequently? I believe that is an error situation where the execution should stop because  the user has an error in the code somewhere. Am I wrong?  Does this happen frequently enough for us to keep the connections alive?

Peque: They should not, for sure. The reason we leave the agent alive after a runtime exception is for debugging purposes (i.e.: the user can create a proxy and see the agent state when the crash occured). Not sure if having the sockets open could help for debugging as well... probably not.

osbrain/tests/test_agent.py
ocaballeror: Yeah, but how are we going to do that? Maybe there's a way to modify the `run` method so we can catch exceptions and shut down the name server (or maybe try to "revive" it), but I can't think of a way to do that.

osbrain/tests/test_agent.py
ocaballeror: Now that you mention it,  it probably would help to keep them open. If you intend to revive the agent after a crash (as long as "crashing" means raising an exception and not any more serious problems), keeping the sockets alive would make things much easier. After all, as long as the agent is not unregistered from the nameserver, calling `ns.shutdown()` will close the socket anyways.

osbrain/tests/test_agent.py
ocaballeror: Makes sense.

osbrain/tests/test_agent.py
ocaballeror: You are right, I tested my example with a longer waiting time and it worked (raised a `ConnectionClosedError` at the end, which was to be expected).

But then again,  your example works better. I'll add it and see if I need to change anything

osbrain/tests/test_agent.py
Peque: @ocaballeror You can keep the test though. Only instead of checking the socket just after the crash, calll `ns.shutdown()` before (you'll need to create the nameserver manually instead of using `nproxy` fixture).

osbrain/tests/test_agent.py
Peque: @ocaballeror Yeah, that is why I suggested to create a separate issue. I am not sure how we could approach this either. Nameserver should *really* never crash (and has never crashed for us when using osBrain). So I do not think it is critical right now. We will need to have a look at it at some point, though. :wink:

ocaballeror: Definitely 👍 

I will remove the test, then. We'll add it again whenever we need to take on that issue.

osbrain/nameserver.py
```diff
@@ -69,6 +69,12 @@ def daemon_shutdown(self):
         """
         self._pyroDaemon.shutdown()
 
+    def raise_exception(self):
+        """
+        Raise an exception (for testing purposes).
+        """
+        raise RuntimeError('User raised an exception!')
+
```
Peque: No need to implement this method now. :wink:

osbrain/tests/test_agent_ipc_sockets.py
```diff
@@ -0,0 +1,74 @@
...
+    nsproxy.shutdown(timeout=1.)
+
+    assert not addr.address.exists()
+
+
+def test_agent_close_ipc_socket_agent_crash():
+    """
+    Check that the socket is closed and the socket file removed after the agent
+    crashes.
+    """
```
Peque: I would rather call this `test_agent_close_ipc_socket_agent_crash_nameserver_shutdown`. And maybe in the docstring: "[...] after the agent crashes and the name server calls for shutdown.".

What you think?

Peque: Maybe we can do the same with the test just above (to specify that the shutdown call is executed from the name server).

osbrain/tests/test_agent_ipc_sockets.py
```diff
@@ -0,0 +1,74 @@
...
+
+def test_agent_close_ipc_socket_nameserver_shutdown(nsproxy):
+    """
+    Check that the socket is closed and the socket file removed after the name
+    server is shut down.
+    """
+    agent = run_agent('agent')
+    addr = agent.bind('PUSH', 'main')
+    nsproxy.shutdown()
+    assert not addr.address.exists()
```
Peque: Not very important but, for style consistency, I would rather keep the same spaces/blank lines in all these tests (i.e.: chose between leaving a blank space between the code and the assertions or not leaving one).

osbrain/nameserver.py
ocaballeror: Good point

Peque:
By the way, it seems the agent kill test is failing. Probably due to the kill method being "asynchronous". Maybe you could try:

```python
from osbrain.helper import agent_dies


def test_xxxxx(nsproxy):
    agent = run_agent('name')
    address = agent.bind('PUSH')
    agent.kill()

    assert agent_dies('name', nsproxy)
    assert not address.address.exists()
```

To see if it helps.

Peque:
Agent crash test seems to be failing as well in Travis: https://travis-ci.org/opensistemas-hub/osbrain/jobs/333633254

You need to have a look at it too before merging. :wink:

osbrain/tests/test_agent_ipc_sockets.py
ocaballeror: I prefer leaving it. I'll check the other functions as well 👍

osbrain/tests/test_agent_ipc_sockets.py
Peque: :joy:

ocaballeror:
I don't want to be *that* guy, but it works on my machine 😃 

I'll make sure to double check next time.

Peque:
@ocaballeror Yeah, you will find a couple of those cases when working with osBrain or any multi-processing/threading asynchronous environment.

Your machine is fast and things (mostly) occur when they are expected. In Travis, many times, that is not the case and weird things happen. It is good because that way we can spot bugs that otherwise would be hard to find and reproduce.

To make sure, we will execute these tests from Travis a couple of times before merging. :wink:

You can also setup Travis in your own fork if you want to have control over it.

ocaballeror:
@Peque Right. I see what you mean. I still don't understand how it could work before, though.

PR #227, opened 2018-01-26
Complete docstrings

### LINKED ISSUE #179: Some documentation docstrings are messed up
Some functions, despite being one liners which make it clear what they do when looking at the code, are missing a docstring. Others, simply have the `Parameters` and `Returns` field in their docstring (with no description of what it does).

When looking at the generated API, this might not look good:
doc_missing

Therefore I think we should document every single method/function regardless of its complexity. Also, for very simple functions such as `Agent.list_timers()`, I think we could get rid of the `Parameters` and `Returns` fields to save some space. What do you think @Peque ?

ocaballeror:
In response to #179 

**Edit**: fixes #179.

### DIFF
--- a/osbrain/address.py
+++ b/osbrain/address.py
@@ -106,11 +106,13 @@ class AgentAddressRole(str):
 
     def twin(self):
         """
+        Get the twin role of the current one. `server` would be the twin
+        of `client` and viceversa.
+
         Returns
         -------
         AgentAddressRole
-            The twin role of the current one; `server` would be the twin
-            of `client` and viceversa.
+            The twin role.
         """
         if self == 'server':
             return self.__class__('client')
@@ -153,32 +155,37 @@ class AgentAddressKind(str):
 
     def zmq(self):
         """
+        Get the equivalent ZeroMQ socket kind.
+
         Returns
         -------
         int
-            The equivalent ZeroMQ socket kind.
         """
         return self.ZMQ_KIND_CONVERSION[self]
 
     def requires_handler(self):
         """
+        Whether the Agent's address kind requires a handler or not.
+        A socket which processes incoming messages would require a
+        handler (i.e. 'REP', 'PULL', 'SUB'...).
+
         Returns
         -------
         bool
-            Whether the Agent's address kind requires a handler or not.
-            A socket which processes incoming messages would require a
-            handler (i.e. 'REP', 'PULL', 'SUB'...).
         """
         return self in self.REQUIRE_HANDLER
 
     def twin(self):
         """
+        Get the twin kind of the current one.
+
+        `REQ` would be the twin of `REP` and viceversa, `PUB` would be the
+        twin of `SUB` and viceversa, etc.
+
         Returns
         -------
         AgentAddressKind
-            The twin kind of the current one; `REQ` would be the twin
-            of `REP` and viceversa, `PUB` would be the twin of `SUB` and
-            viceversa, etc.
+            The twin kind of the current one.
         """
         return self.__class__(self.TWIN[self])
 
@@ -319,13 +326,16 @@ class AgentAddress():
 
     def twin(self):
         """
+        Return the twin address of the current one.
+
+        While the `host` and `port` are kept for the twin, the `kind` and
+        `role` change to their corresponding twins, according to the
+        rules defined in the respective classes.
+
         Returns
         -------
         AgentAddress
-            The twin address of the current one; while the `host` and `port`
-            are kept for the twin, the `kind` and `role` change to their
-            corresponding twins, according to the rules defined in the
-            respective classes.
+            The twin address of the current one.
         """
         kind = self.kind.twin()
         role = self.role.twin()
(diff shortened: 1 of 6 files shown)


osbrain/agent.py
```diff
@@ -1576,6 +1629,9 @@ def shutdown(self):
             self._kill_now = True
 
     def kill(self):
+        """
+        Force shutdown of the agent
```
Peque: .

osbrain/agent.py
```diff
@@ -1548,7 +1598,7 @@ def send_recv(self, address, message):
     @Pyro4.oneway
     def run(self):
         """
-        Run the agent.
+        Begin execution of the agent process and start the main loop.
```
Peque: Not really the begin of the agent process execution. Running the agent means starting the main loop, that is correct though.

osbrain/agent.py
```diff
@@ -1415,6 +1453,9 @@ def _send_address(self, address, message, topic=None):
         self.socket[address].send(message)
 
     def _send_channel(self, channel, message, topic, handler, wait, on_error):
+        """
+        Send a message to a specific channel.
```
Peque: Rather than "to", "through"? (just like in other docstrings below)

osbrain/agent.py
```diff
@@ -1403,6 +1438,9 @@ def send(self, address, message, topic=None, handler=None, wait=None,
         raise NotImplementedError('Unsupported address type %s!' % address)
 
     def _send_address(self, address, message, topic=None):
+        """
+        Send a message to a specific address.
```
Peque: Rather than "to", "through"? (just like in other docstrings below)

osbrain/address.py
```diff
@@ -428,10 +440,12 @@ def __eq__(self, other):
 
     def twin(self):
         """
+        Get the twin channel of the current one.
+
         Returns
         -------
         AgentChannel
-            The twin channel of the current one.
+            The twin channel.
```
Peque: Maybe instead of "Get the twin..." we could say "Return the twin..." and remove the "Returns" section from the docstring. Would apply to other docstrings above this one.

The idea Guillermo proposed was to only use a short description for simple functions/methods (i.e.: no parameters, single return).

On the other hand we lose the return type, but we could use typing for that (after all, the idea is to use typing everywhere for version 1.0). To specify the return type:

```python
def function_or_method(params) -> ReturnType:
    pass
```

What do you think?

ocaballeror: I like having the return type in the docs. Will sphinx be able to read typing and generate similar documentation to what we have right now?

osbrain/agent.py
```diff
@@ -1750,6 +1817,9 @@ def start(self):
                            '\n===============\n'.join(['', status, '']))
 
     def kill(self):
+        """
+        Force kill the agent process
```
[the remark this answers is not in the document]
ocaballeror: I swear I checked every method before committing this time. Some of them must have slipped through 😞

osbrain/address.py
ocaballeror: Yes, we probably should look at this again when we move to typing.

I'll keep it as it is (even if it's redundant), and we'll clean it up later if it's OK with you.

Peque: Yeah, ok for me. Add an issue though, to try typing/Sphinx/napoleon and see if it works well or as expected.

osbrain/agent.py
```diff
@@ -1598,7 +1598,7 @@ def send_recv(self, address, message):
     @Pyro4.oneway
     def run(self):
         """
-        Begin execution of the agent process and start the main loop.
+        Start the main loop of the agent process.
```
Peque: I would not use the word process here. Although it is usually the case, it does not necessarily need to be like that. An agent could run on a thread, or you could even run its main loop from `__main__` if you wanted to. As you see, `Agent` does not inherit from `multiprocessing.Process`.

So I would rather say "Start the main loop of the agent." or "Start the agent's main loop".

osbrain/address.py
```diff
@@ -319,13 +326,16 @@ def __eq__(self, other):
...
+        rules defined in the respective classes.
+
         Returns
         -------
         AgentAddress
-            The twin address of the current one; while the `host` and `port`
-            are kept for the twin, the `kind` and `role` change to their
-            corresponding twins, according to the rules defined in the
-            respective classes.
+            The twin address of the current one
```
Peque: Also... dot. :stuck_out_tongue_winking_eye:

Peque:
Looks good to me. :blush:

osbrain/address.py
Peque: By the way, remember to open an issue for this. :wink:

Peque:
@ocaballeror Remember to push changes to the PR before merging into master. It is safer and that way the PR gets automatically closed once you merge. :wink:

BTW: for future commits try not to use past sentence in the commit messages (i.e.: start with "Add, update, implement..." instead of "Added, updated, implemented..."). I forgot to check that in the review. :sweat_smile:

PR #229, opened 2018-01-30
Added `.configure()` method

### LINKED ISSUE #206: Calling `.each(0, 'x')` in `on_init()` does not work?
This method is executed in `Agent.__init__()`.

Currently, if we execute `.each(0.5, 'x')`, then the agent will crash as the `Agent._loopback()` method dos not support operation when the agent is not running (i.e.: will raise a `NotImplementedError` exception).

We have two options:

- Either we support that operation (i.e.: we return `None` instead of raising an exception). This way messages will be discarded while the agent is not running.
- We create a new method `.before_running()` or `.configure()` which would allow the user to execute code just before entering the `self.loop()` (including `.each(...)` calls). This method should be documented and, maybe, preferred over `.on_init()`, which could be used by advanced users for some special cases.

I think I would rather go with the second option, naming the method `.configure()` or something else but simple/intuitive for the user. @Flood1993 What do you think? Maybe there are other options than those two?

------------

Code to reproduce the error:

```python
import time

from osbrain import Agent
from osbrain import run_agent


class MyAgent(Agent):
    def on_init(self):
        self.each(0.5, 'x')

    def x(self):
        print('asdf')


ns = run_nameserver()
run_agent('asdf', base=MyAgent)
time.sleep(1.5)
ns.shutdown()
```

ocaballeror:
Should fix #206 

I'm not really sure about the warning in the docstrings. I think it's an important thing to note, but I don't  know if you will like how it looks.

### DIFF
--- a/docs/source/timers.rst
+++ b/docs/source/timers.rst
@@ -17,6 +17,11 @@ this, let us modify the :ref:`push_pull` example a bit and make use of the
 Note that if an action takes longer to run than the time available before the
 next execution, the timer will simply fall behind.
 
+.. note:: To be able to start timers on agent initalization, the agent must be
+   already running, which means you cannot do it from ``on_init()``. Instead,
+   you can use ``configure()``, which will be executed when the agent is
+   already running, right before starting the main loop.
+
 
 Delayed actions
 ===============
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -262,6 +262,12 @@ class Agent():
         """
         pass
 
+    def configure(self):
+        """
+        This user-defined method is to be executed right before the main loop.
+        """
+        pass
+
     def _handle_loopback(self, message):
         """
         Handle incoming messages in the loopback socket.
@@ -1608,6 +1614,7 @@ class Agent():
         Start the main loop.
         """
         self.running = True
+        self.configure()
         try:
             self.loop()
         except Exception as error:
--- a/osbrain/tests/test_agent.py
+++ b/osbrain/tests/test_agent.py
@@ -103,6 +103,24 @@ def test_late_runner(nsproxy):
     assert a0.ping() == 'pong'
 
 
+def test_agent_configure(nsproxy):
+    """
+    Tests agent's `configure()` method, which should be able to start timers.
+    """
+    class MyAgent(Agent):
+        def on_init(self):
+            self.x = 0
+
+        def configure(self):
+            self.each(.5, 'incr')
+
+        def incr(self):
+            self.x += 1
+
+    agent = run_agent('agent', base=MyAgent)
+    assert wait_agent_attr(agent, 'x', value=3, timeout=1.2)
+
+
 def test_agent_shutdown(nsproxy):
     """
     An agent must unregister itself before shutting down.
@@ -656,7 +674,7 @@ def test_agent_execute_as_function(nsproxy):
     in the remote agent.
     """
     class EnvironmentAgent(Agent):
-        def configure(self, value):
+        def set_environment(self, value):
             os.environ['__OSBRAIN_TEST'] = value
 
     def name(prefix, suffix='suffix'):
@@ -664,8 +682,8 @@ def test_agent_execute_as_function(nsproxy):
 
     agent0 = run_agent('a0', base=EnvironmentAgent)
     agent1 = run_agent('a1', base=EnvironmentAgent)
-    agent0.configure('0')
-    agent1.configure('1')
+    agent0.set_environment('0')
+    agent1.set_environment('1')
 
     assert agent0.execute_as_function(name, 'p') == 'p0suffix'
     assert agent0.execute_as_function(name, 'p', suffix='s') == 'p0s'


osbrain/tests/test_agent.py
```diff
@@ -644,16 +658,16 @@ def test_agent_execute_as_function(nsproxy):
...
 
     def name(prefix, suffix='suffix'):
         return prefix + os.environ.get('__OSBRAIN_TEST', '') + suffix
 
     agent0 = run_agent('a0', base=EnvironmentAgent)
     agent1 = run_agent('a1', base=EnvironmentAgent)
-    agent0.configure('0')
-    agent1.configure('1')
+    agent0.set_up('0')
+    agent1.set_up('1')
```
Peque: Not very important. For me `.set_environment()` would sound better than `.set_up()`. If you don't agree, leave it as-is.

osbrain/agent.py
```diff
@@ -252,6 +252,18 @@ def __init__(self, name=None, host=None, serializer=None, transport=None,
...
+        is executed when the agent is already running.
+        """
+        pass
+
+    def configure(self):
+        """
+        This user-defined method is to be executed right before the main loop.
+
+        The use of `configure` is preferred over `on_init`, since `configure`
+        is executed when the agent is already running.
```
Peque: I would omit this paragraph here.

osbrain/agent.py
```diff
@@ -252,6 +252,18 @@ def __init__(self, name=None, host=None, serializer=None, transport=None,
     def on_init(self):
         """
         This user-defined method is to be executed after initialization.
+
+        The use of `configure` is preferred over `on_init`, since `configure`
+        is executed when the agent is already running.
```
Peque: Maybe I would omit this paragraph here. It is not really preferred, it is just different, although it is true that most users would be interested in `.configure()` for most use cases.

docs/source/considerations.rst
```diff
@@ -143,6 +143,10 @@ however you may notice some differences:
 #. When setting a handler that is a method already defined in the agent we
    simply pass a string with the method name.
 
+.. note:: The ``configure()`` method is preferred over ``on_init()``, since
+   both will be executed before the main loop, but ``on_init()`` will be
+   called before the agent is running, which can lead to some unexpected errors.
+
```
Peque: Rather than preferred... I would simply state in the note that `on_init()` exists and why is it different.

In example:

"While `configure()` is executed right before the main loop (i.e.: when the agent is already running), we might want to use the lower-level `on_init()` method, which is executed right after `Agent`'s initialization."

You could also remove the paragraph all together and let the user know only about the `.configure()` method... We might even deprecate/remove `.on_init()` in the future if we find out it is not necessary.

osbrain/tests/test_agent.py
```diff
@@ -91,6 +91,20 @@ def run(self):
...
+        def configure(self):
+            self.x = 0
+            self.each(.5, 'incr')
+
+        def incr(self):
+            self.x += 1
+
+    agent = run_agent('agent', base=MyAgent)
+    time.sleep(1.2)
+    assert agent.get_attr('x') == 3
```
Peque: Maybe we need to rethink all this...

I would have said:

```python
assert wait_agent_attr(agent, name='x', value=3, timeout=1.2)
```

But this can break as there is a race condition: we might check the attribute before `.configure()` is executed. Maybe it makes sense to have both methods: `on_init()` for variables/attributes/parameters, and `.configure()` for starting timers, etc..

Not sure about this, will have to think about it.

docs/source/considerations.rst
ocaballeror: I'd rather keep the note for now. As you said, we can remove `on_init` in the future if turns out not to be necessary after all.

osbrain/tests/test_agent.py
ocaballeror: Didn't look too much into this test. I just used the first similar word to "configure" I could come up with 😄

osbrain/tests/test_agent.py
ocaballeror: That would make sense. However, although possible, I still don't think we will ever run into that race condition.

Though if you think we still should worry about it, using `on_init()` should be fine.

Peque: We should always worry about possible race conditions. Sooner or later they bite you and they are a pain in the $%&@# to debug... :wink:

Maybe we could add another attribute `configured` which is set to true after configuration. Maybe we should wait for that variable to be true when we execute `run_agent` rather than waiting for `running` to be true.

What do you think?

Peque: Yeah, it would be okay. Harder for the user only. But if we do not find a better alternative it might be the best thing to do anyway.

If we go this way we should leave the `.on_init()` in the example and docs just like before and maybe add a paragraph in the Timers section to tell the user about the existence of the `.configure()` method.

ocaballeror:
There is one slight problem with all of this. By making the `name` argument optional, the tests will report a style error, because now both  `Agent.__init___()` and `run_agent()`  are "too complex".

What should we do about this? `tox` will mark our entire test suite as "failing".

docs/source/considerations.rst
```diff
@@ -133,8 +133,8 @@ however you may notice some differences:
    instead. In this case, this class is named ``Greeter``.
 #. The ``Greeter`` class implements two methods:
 
-   #. ``on_init()``: which is executed on initialization and will, in this
-      case, simply bind a ``'PUSH'`` communication channel.
+   #. ``on_init()``: which is executed before the main loop starts, and will,
+      in this case, simply bind a ``'PUSH'`` communication channel.
```
Peque: You need to revert the changes in the whole paragraph. I think you just changed back to "on_init", but forgot the description. :wink:

docs/source/considerations.rst
Peque: I would remove this paragraph now. Nothing changed and `on_init` should still be used for initial attributes setting and socket binding. We added a new method, but that is explained in another section (in the timers section).

docs/source/timers.rst
```diff
@@ -17,6 +17,11 @@ this, let us modify the :ref:`push_pull` example a bit and make use of the
 Note that if an action takes longer to run than the time available before the
 next execution, the timer will simply fall behind.
 
+.. note:: To be able to run the ``each()`` method, the agent must be already
+   running, which means you cannot call it from ``on_init()``. Instead, you can
+   use ``configure()``, which will be executed when the agent is already
+   running, right before starting the main loop.
```
Peque: I would rather say: "To be able to start timers on agent initialization, the agent must be[...]".

examples/push_pull_inherit.py
```diff
@@ -5,7 +5,7 @@
 
 
 class Greeter(Agent):
-    def on_init(self):
+    def configure(self):
         self.bind('PUSH', alias='main')
```
Peque: I would revert this. `configured` is not yet recommended over `on_init` (they both serve for different purposes).

osbrain/tests/test_agent.py
Peque: If you are okay then I would rather use the `on_init()` here for setting `x` and then use the `assert wait_agent_attr()`.

docs/source/considerations.rst
ocaballeror: Got me there 😀 .

osbrain/tests/test_agent.py
```diff
@@ -91,6 +91,22 @@ def run(self):
     assert a0.ping() == 'pong'
 
 
+def test_agent_configure(nsproxy):
+    class MyAgent(Agent):
+        def on_init(self):
+            self.x = 0
+
+        def configure(self):
+            self.x = 0
```
Peque: Duplicated `self.x = 0`.

docs/source/considerations.rst
```diff
@@ -133,7 +133,7 @@ however you may notice some differences:
    instead. In this case, this class is named ``Greeter``.
 #. The ``Greeter`` class implements two methods:
 
-   #. ``on_init()``: which is executed on initialization and will, in this
+   #. ``on_init()``: which is executed on initialization, and will, in this
```
Peque: No need for that comma after "[...] on initialization, [...]".

osbrain/tests/test_agent.py
```diff
@@ -91,6 +91,21 @@ def run(self):
     assert a0.ping() == 'pong'
 
 
+def test_agent_configure(nsproxy):
```
Peque: Maybe a short docstring for the test and you are good to go! :blush:

(Even if it is a simple "Test agent's `configure()` method, which should be able to start timers.")

ocaballeror:
Closed PR with commit: `1f5ffb0ed36d0bb6d3f78a3a2c191e03963b5b62`

PR #231, opened 2018-01-30
Auto generate names for agents

### LINKED ISSUE #223: Allow running agents without alias
If an alias is missing, maybe a random alias should be generated.

ocaballeror:
Fixes #223 

I  was unsure how to test it, though. This is the best solution I could come up with.

### DIFF
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -205,10 +205,13 @@ class Agent():
     running : bool
         Set to `True` if the agent is running (executing the main loop).
     """
-    def __init__(self, name=None, host=None, serializer=None, transport=None,
+    def __init__(self, name='', host=None, serializer=None, transport=None,
                  attributes=None):
         self.uuid = unique_identifier()
         self.name = name
+        if not self.name:
+            self.name = self.uuid.decode()
+
         self.host = host
         if not self.host:
             self.host = '127.0.0.1'
@@ -1724,7 +1727,7 @@ class AgentProcess(multiprocessing.Process):
     Agent class. Instances of an Agent are system processes which
     can be run independently.
     """
-    def __init__(self, name, nsaddr=None, addr=None, serializer=None,
+    def __init__(self, name='', nsaddr=None, addr=None, serializer=None,
                  transport=None, base=Agent, attributes=None):
         super().__init__()
         self.name = name
@@ -1759,6 +1762,7 @@ class AgentProcess(multiprocessing.Process):
             self.queue.put(format_exception())
             return
 
+        self.name = self.agent.name
         uri = self._daemon.register(self.agent)
         try:
             ns.register(self.name, uri, safe=True)
@@ -1768,7 +1772,7 @@ class AgentProcess(multiprocessing.Process):
         finally:
             ns.release()
 
-        self.queue.put('STARTED')
+        self.queue.put('STARTED:'+self.name)
 
         self._daemon.requestLoop(lambda: not self.shutdown_event.is_set())
         self._daemon.unregister(self.agent)
@@ -1811,10 +1815,11 @@ class AgentProcess(multiprocessing.Process):
         """
         super().start()
         status = self.queue.get()
-        if status == 'STARTED':
-            return
-        raise RuntimeError('An error occured while creating the daemon!' +
-                           '\n===============\n'.join(['', status, '']))
+        if not status.startswith('STARTED'):
+            raise RuntimeError('An error occured while creating the daemon!' +
+                               '\n===============\n'.join(['', status, '']))
+
+        return status.partition(':')[-1]
 
     def kill(self):
         """
@@ -1832,7 +1837,7 @@ class AgentProcess(multiprocessing.Process):
         self.kill()
 
 
-def run_agent(name, nsaddr=None, addr=None, base=Agent, serializer=None,
+def run_agent(name='', nsaddr=None, addr=None, base=Agent, serializer=None,
               transport=None, safe=None, attributes=None):
     """
     Ease the agent creation process.
@@ -1842,7 +1847,7 @@ def run_agent(name, nsaddr=None, addr=None, base=Agent, serializer=None,
 
     Parameters
     ----------
-    name : str
+    name : str, default is ''
         Agent name or alias.
     nsaddr : SocketAddress, default is None
         Name server address.
@@ -1862,10 +1867,12 @@ def run_agent(name, nsaddr=None, addr=None, base=Agent, serializer=None,
     """
     if not nsaddr:
         nsaddr = os.environ.get('OSBRAIN_NAMESERVER_ADDRESS')
-    AgentProcess(name, nsaddr=nsaddr, addr=addr, base=base,
-                 serializer=serializer, transport=transport,
-                 attributes=attributes).start()
-    proxy = Proxy(name, nsaddr, safe=safe)
+    agent = AgentProcess(name=name, nsaddr=nsaddr, addr=addr, base=base,
+                         serializer=serializer, transport=transport,
+                         attributes=attributes)
+    agent_name = agent.start()
+
+    proxy = Proxy(agent_name, nsaddr, safe=safe)
     proxy.run()
     proxy.wait_for_running()
 
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -244,13 +244,17 @@ class Agent():
                   handler=self._handle_loopback_safe, transport='inproc',
                   serializer='pickle')
 
-        if attributes:
-            for key, value in attributes.items():
-                if hasattr(self, key):
-                    raise KeyError('Agent already has "%s" attribute!' % key)
-                setattr(self, key, value)
-
+        self._set_attributes(attributes)
         self.on_init()
+
+    def _set_attributes(self, attributes):
+        if not attributes:
+            return
+
+        for key, value in attributes.items():
+            if hasattr(self, key):
+                raise KeyError('Agent already has "%s" attribute!' % key)
+            setattr(self, key, value)
 
     def on_init(self):
         """
--- a/osbrain/tests/test_agent.py
+++ b/osbrain/tests/test_agent.py
@@ -39,6 +39,18 @@ def test_agent_uuid():
     assert all(isinstance(identifier, bytes) for identifier in bunch)
 
 
+def test_agent_auto_generated_name(nsproxy):
+    """
+    If an agent is initialized without a name, a globally unique one must be
+    auto generated for it.
+    """
+    count = 10
+    names = set([run_agent().get_attr('name') for x in range(count)])
+    assert all(isinstance(name, str) for name in names)
+    assert len(names) == count
+    assert names == set(nsproxy.agents())
+
+
 def test_early_agent_proxy(nsproxy):
     """
     It must be possible to create a Proxy when the registration of the new


osbrain/agent.py
```diff
@@ -1668,10 +1668,13 @@ class AgentProcess(multiprocessing.Process):
...
     """
-    def __init__(self, name, nsaddr=None, addr=None, serializer=None,
+    def __init__(self, name=None, nsaddr=None, addr=None, serializer=None,
                  transport=None, base=Agent, attributes=None):
         super().__init__()
-        self.name = name
+        if name:
+            self.name = name
+        else:
+            self.name = unique_identifier().decode()
```
Peque: Maybe better:

```python
if name is None:
    name = unique_identifier().decode()
```

Anyway this should not be here, as you are not reusing the Agent's UUID. Was not that the idea?

osbrain/tests/test_agent.py
```diff
@@ -39,6 +39,17 @@ def test_agent_uuid():
...
+def test_agent_auto_generated_name(nsproxy):
+    """
+    If an agent is initialized without a name, a globally unique one must be
+    auto generated for it.
+    """
+    agent = run_agent()
+    name = agent.get_attr('name')
+    assert isinstance(name, str)
+    assert name != ''
+
```
Peque: Yeah, this is a minimal check. But I would rather do something like:

```python
def test_xxx(nsproxy):
    N = 10
    names = [run_agent().get_attr('name') for x in range(N)]
    assert all(isinstance(name, str) for name in names)
    assert len(set(names)) == N
    assert set(names) == set(nsproxy.agents())
```

osbrain/agent.py
ocaballeror: Maybe I got this wrong, but this is  `AgentProcess`, which doesn't have a UUID attribute. It's the `Agent` class that has it, so we should probably generate a new UUID and use that instead.

osbrain/tests/test_agent.py
ocaballeror: That's one way to check they don't collide. Still not perfect, though, but I can't think of anything better.

osbrain/agent.py
Peque: Well, you suggested reusing it! :joy:

Anyway I like that idea, and it definitely can be done:

- The `Agent()` class should allow for an empty `name` attribute. If `name is None` then we assign to it the string representation of the Agent's UUID.
- In `AgentProcess`'s `run()` method we can get the `Agent` name attribute just before registering it in the daemon. Then, for registering in the name server we use that name instead of `self.name`.
- Now we need to tell the parent the name that was used. We can pass that information in `self.queue.put('STARTED:name')` (for example).
- Then the `AgentProcess`'s `start()` method should return the name that was used.
- `run_agent()` then knows the name as well and can create the proxy using the returned name instead of the passed `name` parameter.

ocaballeror: You're right. This seems to work so far.

ocaballeror:
There is one slight problem with all of this. By making the name argument optional, the tests will report a style error, because now both Agent.__init___() and run_agent() are "too complex".

What should we do about this? tox will mark our entire test suite as "failing".

Peque:
If it is too complex we need to simplify! :stuck_out_tongue_winking_eye: 

I will leave some comments on the code (will be easier).

osbrain/agent.py
```diff
@@ -1668,10 +1672,13 @@ class AgentProcess(multiprocessing.Process):
...
     """
-    def __init__(self, name, nsaddr=None, addr=None, serializer=None,
+    def __init__(self, name=None, nsaddr=None, addr=None, serializer=None,
                  transport=None, base=Agent, attributes=None):
         super().__init__()
-        self.name = name
+        if name is None:
+            self.name = ''
+        else:
+            self.name = name
```
Peque: You can revert this back to `self.name = name`. No need for this `if-else` I think.

osbrain/agent.py
```diff
@@ -1700,6 +1707,8 @@ def run(self):
             self.queue.put(format_exception())
             return
 
+        if not self.name:
+            self.name = self.agent.name
```
Peque: No need for this `if` either. Just `self.name = self.agent.name`, always.

osbrain/agent.py
```diff
@@ -209,6 +210,9 @@ def __init__(self, name=None, host=None, serializer=None, transport=None,
                  attributes=None):
         self.uuid = unique_identifier()
         self.name = name
+        if not self.name:
+            self.name = self.uuid.decode()
+
```
Peque: It seems I cannot review parts of the code that were not changed... :sweat_smile: 

Anyway... Here you will probably want to move the `if attributes:` section to a new method `_set_attributes(self, attributes):`.

osbrain/agent.py
ocaballeror: I needed to do it this way. Setting `self.name` to None raises an `AssertionError` from the multiprocessing library. Apparently, `name` must be a string.

osbrain/agent.py
ocaballeror: I wanted to specify that these 2 lines are specifically added to handle the cases where the `AgentProcess` doesn't have a name already. If it did, `self.name = self.agent.name` would look a bit weird, since we just passed `self.name` to the `Agent`'s constructor a few lines above.

Maybe I could remove the `if``and add a comment if you think it looks better.

osbrain/agent.py
Peque: Yeah, forgot `name` is an actual parameter of the `multiptrocessing.Process` class.

Maybe you can just change the defaults to `''` instead of `None`. Would that work?

osbrain/agent.py
ocaballeror: Why exactly? Is it just a matter of style?

osbrain/agent.py
Peque: Without the `if` clause we reduce cyclomatic complexity, which is always good. Also, we do not know if the name of the agent was changed since we passed it as parameter on initialization, so executing `self.name = self.agent.name` is a way to ensure the process name is the agent's name, which is what we want.

The idea here is: if this code:

```python
if condition:
    foo()
```

Gives the same exact results as this one:

```python
foo()
```

Then we would normally prefer the latter, as it is objectively less complex. Unless `foo()` is really a complex task and we want to avoid its execution if possible (that is not really the case).

Feel free to add a comment if you think it is easier to understand it that way.

osbrain/agent.py
Peque: `__init__` became too complex, which usually tells us that we should probably split this method/function in smaller (less complex) parts. This eases the process of testing and code reusing.

I suggested the `if attributes` section as it seems it has its own meaning and adds 3 to the cyclomatic complexity. So that is a perfect candidate for a split.

You can even add it as a function rather than a method. In example: `def add_dictionary_attributes(target, attributes):`. As this could be applied to any object, not just `Agent`.

Peque: PS: this refactoring (i.e.: moving that section into a new function) should be a separate commit. Keep it in this PR though.

Peque: PSS: :joy: Even if that section is covered, you might want to add simple tests to verify the new function.

osbrain/agent.py
ocaballeror: Yeah, that's quite simpler, actually. And it seems to work just fine.

osbrain/agent.py
ocaballeror: OK, then. I agree that is a separate issue and it should be kept in a separate commit.

ocaballeror: Also, I just tested it out and now the tests seem to pass 👍 .

osbrain/agent.py
```diff
@@ -1744,8 +1751,12 @@ def _teardown(self):
...
         super().start()
         status = self.queue.get()
-        if status == 'STARTED':
-            return
+        if status.startswith('STARTED'):
+            match = re.match(r'STARTED: (.*)', status)
+            name = ''
+            if match:
+                name = match.group(1)
+            return name
```
Peque: No need to use regular expressions here:

```python
return status.partition(':')[-1]
```

We could also say:

```python
if not status.startswith('STARTED:'):
    raise [...]
return [...]
```

Looks nicer to me in this case.

osbrain/agent.py
```diff
@@ -1772,7 +1783,7 @@ def run_agent(name, nsaddr=None, addr=None, base=Agent, serializer=None,
 
     Parameters
     ----------
-    name : str
+    name : str, default is None
```
Peque: Default is `''`. :wink:

We might remove all this redundancy when we integrate typing. For now we will keep it for style consistency (this is the NumPy/Pandas/Scipy style, which we initially chose).

osbrain/agent.py
```diff
@@ -1792,10 +1803,16 @@ def run_agent(name, nsaddr=None, addr=None, base=Agent, serializer=None,
...
-    proxy = Proxy(name, nsaddr, safe=safe)
+    agent = AgentProcess(name=name, nsaddr=nsaddr, addr=addr, base=base,
+                         serializer=serializer, transport=transport,
+                         attributes=attributes)
+    agent_name = agent.start()
+    if name:
+        agent_name = name
+
+    print('AGENT NAME: ', name)
+
```
Peque: I think you forgot this print and the `if name:`.

osbrain/agent.py
ocaballeror: I've gotten too used to parsing everything with regexes. Too much use of Unix shells I guess 😜

osbrain/agent.py
ocaballeror: Didn't forget that `if`. This one actually makes sense.

I did miss the `print`, though.

Peque: Does the result change if you remove that `if`? (or does any test break?)

ocaballeror: It does change. This one is necessary, so the tests will probably fail if it's removed.

Peque: Probably? Or for sure? :wink:

ocaballeror: Just tried it out, and they fail. So yeah, for sure 😉

Peque: Which tests fail?

ocaballeror: `test_agent_auto_generated_name`,  which is the one I added for this.

This `if` avoids overriding the `agent_name` with the `name` passed as a parameter,  which is what we want to do if we didn't receive any parameters (basically the whole point of this PR). So yeah, it's important to keep it.

Peque: I find that weird... Can you show me the `pytest` output so see where it fails? (or upload it to see the Travis report)

ocaballeror: How is that weird? I don't understand what you see wrong with these lines.
In the previous statement, we catch  the return value of `agent.start()`, but if a `name` was passed as a parameter, we should use that instead to create the `Proxy` on the next line. Otherwise we won't be able to get a `Proxy` from the nameserver because we will be querying for the wrong agent name.

Anyway, here is the full output from pytest (it times out while creating the object).
```
============================================== test session starts ==============================================
platform linux -- Python 3.6.4, pytest-3.3.2, py-1.5.2, pluggy-0.6.0 -- /usr/bin/python
cachedir: .cache
rootdir: /home/ocaballero/Documents/osbrain, inifile: setup.cfg
plugins: xdist-1.22.0, rerunfailures-4.0, pep8-1.0.6, mccabe-0.1, forked-0.2, flakes-2.0.0, cov-2.5.1
collected 258 items                                                                                             

osbrain/tests/test_agent.py::test_agent_auto_generated_name FAILED                                        [100%]

=================================================== FAILURES ====================================================
________________________________________ test_agent_auto_generated_name _________________________________________

nsproxy = <osbrain.proxy.NSProxy at 0x7faa3fc26940; not connected; for PYRONAME:Pyro.NameServer@127.0.0.1:11521>

    def test_agent_auto_generated_name(nsproxy):
        """
        If an agent is initialized without a name, a globally unique one must be
        auto generated for it.
        """
        count = 10
>       names = set([run_agent().get_attr('name') for x in range(count)])

osbrain/tests/test_agent.py:48: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
osbrain/tests/test_agent.py:48: in <listcomp>
    names = set([run_agent().get_attr('name') for x in range(count)])
osbrain/agent.py:1810: in run_agent
    proxy = Proxy(agent_name, nsaddr, safe=safe)
osbrain/proxy.py:80: in __init__
    while not self._ready_or_timeout(time0, timeout):
osbrain/proxy.py:96: in _ready_or_timeout
    self.unsafe.ping()
osbrain/proxy.py:154: in __getattr__
    return super().__getattr__(name)
/usr/lib/python3.6/site-packages/Pyro4/core.py:273: in __getattr__
    self._pyroGetMetadata()
/usr/lib/python3.6/site-packages/Pyro4/core.py:600: in _pyroGetMetadata
    self.__pyroCreateConnection()
/usr/lib/python3.6/site-packages/Pyro4/core.py:499: in __pyroCreateConnection
    uri = _resolve(self._pyroUri, self._pyroHmacKey)
/usr/lib/python3.6/site-packages/Pyro4/core.py:1853: in _resolve
    return nameserver.lookup(uri.object)
/usr/lib/python3.6/site-packages/Pyro4/core.py:187: in __call__
    return self.__send(self.__name, args, kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <Pyro4.core.Proxy at 0x7faa44621f28; not connected; for PYRO:Pyro.NameServer@127.0.0.1:11521>
methodname = 'lookup', vargs = ('@127.0.0.1:11521',), kwargs = {}, flags = 0, objectId = 'Pyro.NameServer'

    def _pyroInvoke(self, methodname, vargs, kwargs, flags=0, objectId=None):
        """perform the remote method call communication"""
        current_context.response_annotations = {}
        with self.__pyroConnLock:
            if self._pyroConnection is None:
                self.__pyroCreateConnection()
            serializer = util.get_serializer(self._pyroSerializer or config.SERIALIZER)
            objectId = objectId or self._pyroConnection.objectId
            annotations = self.__annotations()
            if vargs and isinstance(vargs[0], SerializedBlob):
                # special serialization of a 'blob' that stays serialized
                data, compressed, flags = self.__serializeBlobArgs(vargs, kwargs, annotations, flags, objectId, methodname, serializer)
            else:
                # normal serialization of the remote call
                data, compressed = serializer.serializeCall(objectId, methodname, vargs, kwargs, compress=config.COMPRESSION)
            if compressed:
                flags |= message.FLAGS_COMPRESSED
            if methodname in self._pyroOneway:
                flags |= message.FLAGS_ONEWAY
            self._pyroSeq = (self._pyroSeq + 1) & 0xffff
            msg = message.Message(message.MSG_INVOKE, data, serializer.serializer_id, flags, self._pyroSeq,
                                  annotations=annotations, hmac_key=self._pyroHmacKey)
            if config.LOGWIRE:
                _log_wiredata(log, "proxy wiredata sending", msg)
            try:
                self._pyroConnection.send(msg.to_bytes())
                del msg  # invite GC to collect the object, don't wait for out-of-scope
                if flags & message.FLAGS_ONEWAY:
                    return None  # oneway call, no response data
                else:
                    msg = message.Message.recv(self._pyroConnection, [message.MSG_RESULT], hmac_key=self._pyroHmacKey)
                    if config.LOGWIRE:
                        _log_wiredata(log, "proxy wiredata received", msg)
                    self.__pyroCheckSequence(msg.seq)
                    if msg.serializer_id != serializer.serializer_id:
                        error = "invalid serializer in response: %d" % msg.serializer_id
                        log.error(error)
                        raise errors.SerializeError(error)
                    if msg.annotations:
                        current_context.response_annotations = msg.annotations
                        self._pyroResponseAnnotations(msg.annotations, msg.type)
                    if self._pyroRawWireResponse:
                        msg.decompress_if_needed()
                        return msg
                    data = serializer.deserializeData(msg.data, compressed=msg.flags & message.FLAGS_COMPRESSED)
                    if msg.flags & message.FLAGS_ITEMSTREAMRESULT:
                        streamId = msg.annotations.get("STRM", b"").decode()
                        if not streamId:
                            raise errors.ProtocolError("result of call is an iterator, but the server is not configured to allow streaming")
                        return _StreamResultIterator(streamId, self)
                    if msg.flags & message.FLAGS_EXCEPTION:
                        if sys.platform == "cli":
                            util.fixIronPythonExceptionForPickle(data, False)
>                       raise data  # if you see this in your traceback, you should probably inspect the remote traceback as well
E                       Pyro4.errors.NamingError: unknown name: @127.0.0.1:11521

/usr/lib/python3.6/site-packages/Pyro4/core.py:472: NamingError
--------------------------------------------- Captured stdout setup ---------------------------------------------
Broadcast server running on 0.0.0.0:9091
NS running on 127.0.0.1:11521 (127.0.0.1)
URI = PYRO:Pyro.NameServer@127.0.0.1:11521
============================================= 257 tests deselected ==============================================
=================================== 1 failed, 257 deselected in 3.69 seconds ====================================
```

Peque: From the traceback (seeing that your `proxy = Proxy(..)` is in line 1810), maybe you thought I was referring to change this:

```python
agent_name = agent.start()
if name:
    agent_name = name

proxy = Proxy(...)
```

Into this:

```python
agent_name = agent.start()
agent_name = name

proxy = Proxy(...)
```

But that does not make sense, as the first assignment would be unnecessary. I was referring to remove the if-clause all together:

```python
agent_name = agent.start()

proxy = Proxy(...)
```

ocaballeror: Yeah, I thought you meant that all along, hence my confusion 😁 

Removing the entire clause does work. When I wrote that I thought the agent name would only be returned if we passed nothing as  parameter. Now it makes sense to remove it 👍

Peque: Candidate for the longest 2-lines discussion in GitHub. :joy:

Issue #200, opened 2017-10-23
Document agent attributes


Peque:
`name`, `uuid`, ...

Any common/default agent attribute.

Maybe simply expand/complete https://osbrain.readthedocs.io/en/stable/api/agent.html.

What about #205?

ocaballeror:
About to start working on this.

About the other issue you referenced, I don't think it's necessary, but if we are going to do this, we might as well document the private attributes as  well. It makes it easier for anyone looking to contribute to the project.

Peque:
@ocaballeror Great. I think we can finish this first. Then we take on #200 (and #225, which is similar).

ocaballeror:
Closed by #232

PR #232, opened 2018-02-02
Add documentation for Agent attributes

ocaballeror:
In response to #200

### DIFF
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -185,8 +185,14 @@ class Agent():
     ----------
     name : str
         Name of the agent.
+    uuid : bytes
+        Globally unique identifer for the agent.
     host : str
         Host address where the agent is binding to.
+    serializer : str
+        Default agent serialization format.
+    transport : str, AgentAddressTransport, default is None
+        Default agent transport protocol.
     socket : dict
         A dictionary in which the key is the address or the alias and the
         value is the actual socket.
@@ -196,6 +202,10 @@ class Agent():
     handler : dict
         A dictionary in which the key is the socket and the values are the
         handlers for each socket.
+    context : zmq.Context
+        ZMQ context to create ZMQ socket objects.
+    poller : zmq.Poller
+        ZMQ poller to wait for incoming data from sockets.
     poll_timeout : int
         Polling timeout, in milliseconds. After this timeout, if no message
         is received, the agent executes de `idle()` method before going back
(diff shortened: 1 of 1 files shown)


osbrain/agent.py
```diff
@@ -204,6 +214,19 @@ class Agent():
         When set to `True`, the agent will continue executing the main loop.
     running : bool
         Set to `True` if the agent is running (executing the main loop).
+    _async_req_uuid : dict
+        Stores a unique identifier for every REQ or PULL connection.
+    _async_req_handler : dict
+        Stores the handler for every active REQ or PULL connection.
```
Peque: I think these `_async_req_*` are incorrect, but would need to have a look at the code to know what they are exactly doing. Maybe it is not the best name, as they were probably initially created for ASYNC-REQ channels but are now used for SYNC-PUB channels as well...

osbrain/agent.py
```diff
@@ -204,6 +214,19 @@ class Agent():
...
+    _kill_now : bool
+        Used internally to synchronously shut down the agent.
```
Peque: I'd say "During shutdown, this attribute is set for the agent to be killed.`.

osbrain/agent.py
```diff
@@ -204,6 +214,19 @@ class Agent():
...
+    _DEBUG : bool
+        Whether to print debug level messages.
+    _pending_requests : dict
+        Stores the handler for every REQ message sent while waiting for the
+        reply.
```
Peque: I'd say "Stores pending (without reply) asynchronous requests. The asynchronous request UUID is used as key and the handler is set as value." Note that "asynchronous requests" are not the same as REQ messages. Need to add documentation on channels, which is currently missing... :sweat:

osbrain/agent.py
```diff
@@ -196,6 +202,10 @@ class Agent():
     handler : dict
         A dictionary in which the key is the socket and the values are the
         handlers for each socket.
+    context : zmq.Context
+        Used internally to create ZMQ socket objects.
+    poller : zmq.Poller
+        Used internally to wait for incoming data from sockets.
```
Peque: These will be made private soon, so maybe no need to add the "Used internally". In example:

- ZMQ context to create ZMQ socket objects.
- ZMQ poller to wait for incoming data from sockets.

ocaballeror: Yeah, I planned to change those  descriptions when the attributes are made private.

osbrain/agent.py
ocaballeror: This is all I could guess from reading the code. If this is not what they are really used for, I will need a more detailed description from you 😄

osbrain/agent.py
ocaballeror: That would definitely be helpful for some of us 👍 

There are still a couple of things that are not fully clear to me.

osbrain/agent.py
Peque: Maybe:

- Stores the UUIDs of the asynchronous request sockets (used in communication channels).
- Stores the handler for every asynchronous request sockets (used in communication channels).

osbrain/agent.py
Peque: Yeah, will try to do that today/tomorrow.

osbrain/agent.py
ocaballeror: Yeah, I'll definitely need to read a bit more on how channels work here.

PR #234, opened 2018-02-05
Make some methods private

### LINKED ISSUE #225: Revise "private" methods
We should probably go through the entire API and see which methods should be "private" (i.e. have an underscore before the name), since we have found some of those in the last few days.

For example `get_unique_external_zmq_sockets`

ocaballeror:
I went through the entire codebase looking for methods that should probably be private. These are the ones I found.

Fixes #225.

### DIFF
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -501,7 +501,7 @@ class Agent():
         """
         level = LogLevel(level)
         message = '[%s] (%s): %s' % (datetime.utcnow(), self.name, message)
-        if self.registered(logger):
+        if self._registered(logger):
             logger_kind = AgentAddressKind(self._address[logger].kind)
             assert logger_kind == 'PUB', \
                 'Logger must use publisher-subscriber pattern!'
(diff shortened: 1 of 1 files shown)


osbrain/logging.py
```diff
@@ -30,14 +30,14 @@ def on_init(self):
...
-            'ERROR': self.log_handler,
-            'DEBUG': self.log_handler
+            'INFO': self._log_handler,
+            'WARNING': self._log_handler,
+            'ERROR': self._log_handler,
+            'DEBUG': self._log_handler
         }
         self.bind('SUB', 'sub', handlers)
 
-    def log_handler(self, message, topic):
```
Peque: I would not set `log_handler` as private. It is more like a dummy method that the user would probably want to override anyway (unless until we implement better logging...).

osbrain/agent.py
```diff
@@ -1824,7 +1824,7 @@ def kill(self):
         if self._daemon:
             self._daemon.shutdown()
 
-    def sigint_handler(self, signal, frame):
+    def _sigint_handler(self, signal, frame):
         """
         Handle interruption signals.
         """
```
Peque: Other candidate methods to set as private:

- `Agent.register`
- `Agent.registered`

ocaballeror: Missed those 👍

PR #236, opened 2018-02-05
[WIP] Include Python 3.7 in tests

Peque:
For now this is just to see if everything seems to work fine with 3.7 development branch.


ocaballeror:
Is it really necessary to support a development version of python? The final release is not supposed to be released until June this year.

There probably won't be any incompatibilities, there's no need to rush this 😁 .

Peque:
@ocaballeror Sure, it is marked as WIP and assigned to milestone 0.7.0. We will not be merging this until 3.7 is out, but wanted to check compatibility before final release.

Although unlikely, major releases come with many changes and might break something. Better to spot it early and report it if that is the case. The fact that we are using `3.4.4` in Travis is actually an example of `3.4.3` being broken (in this case a broken *minor* release). Checking all minor releases before they are out is not worth it, but major releases are worth it in my opinion (fortunately they do not occur very often).

.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: python
 
 python:
+  - 3.7
```
irmen: travis doens't have 3.7 release version yet. You can try "3.7-dev"  which runs an old beta version, which may be good enough

irmen:
you'll have to use "3.7-dev" for travis because they only have an old 3.7 beta version available right now

Peque:
@irmen Thanks! :blush:

Yeah, we used that during the 3.7 beta to make sure the new version did not break anything. Not that we are in a hurry, so we will probably wait until `3.7` is available in Travis (hope it wont take too long!). :joy:

By the way, if you want to use 3.7 now I think you can, but you need to use `sudo: true` and `dist: xenial` in your `travis.yml` file. A bit of a hack... :sweat_smile:

Peque:
@ocaballeror It seems they finally made Xenial available in Travis. It was about time! :joy:

.travis.yml
Peque: It finally has this version (need to use `dist: xenial` though). :blush:

Issue #205, opened 2017-10-30
Hide all internal attributes


Peque:
In example: `.socket` ---> `._socket` or `._osbrain_socket`.

What about #200 then?

PR #238, opened 2018-02-05
Hide all internal attributes

ocaballeror:
#205.

These are all the private attributes I could find. If you think any of them should be public, we can peacefully discuss it 😁

### DIFF
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -185,35 +185,36 @@ class Agent():
     ----------
     name : str
         Name of the agent.
-    uuid : bytes
-        Globally unique identifer for the agent.
-    host : str
+
+    _host : str
         Host address where the agent is binding to.
-    serializer : str
+    _uuid : bytes
+        Globally unique identifer for the agent.
+    _running : bool
+        Set to `True` if the agent is running (executing the main loop).
+    _serializer : str
         Default agent serialization format.
-    transport : str, AgentAddressTransport, default is None
+    _transport : str, AgentAddressTransport, default is None
         Default agent transport protocol.
-    socket : dict
+    _socket : dict
         A dictionary in which the key is the address or the alias and the
         value is the actual socket.
-    address : dict
+    _address : dict
         A dictionary in which the key is the address or the alias and the
         value is the actual address.
-    handler : dict
+    _handler : dict
         A dictionary in which the key is the socket and the values are the
         handlers for each socket.
-    context : zmq.Context
+    _context : zmq.Context
         ZMQ context to create ZMQ socket objects.
-    poller : zmq.Poller
+    _poller : zmq.Poller
         ZMQ poller to wait for incoming data from sockets.
-    poll_timeout : int
+    _poll_timeout : int
         Polling timeout, in milliseconds. After this timeout, if no message
         is received, the agent executes de `idle()` method before going back
         to polling.
-    keep_alive : bool
+    _keep_alive : bool
         When set to `True`, the agent will continue executing the main loop.
-    running : bool
-        Set to `True` if the agent is running (executing the main loop).
     _async_req_uuid : dict
         Stores the UUIDs of the asynchronous request sockets (used in
         communication channels).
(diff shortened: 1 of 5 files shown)


osbrain/agent.py
```diff
@@ -193,20 +193,20 @@ class Agent():
         Default agent serialization format.
     transport : str, AgentAddressTransport, default is None
         Default agent transport protocol.
```
Peque: You think name, UUID, host, serializer and transport should not be hidden?

osbrain/agent.py
```diff
@@ -193,20 +193,20 @@ class Agent():
...
+    _context : zmq.Context
         ZMQ context to create ZMQ socket objects.
-    poller : zmq.Poller
+    _poller : zmq.Poller
         ZMQ poller to wait for incoming data from sockets.
-    poll_timeout : int
+    _poll_timeout : int
         Polling timeout, in milliseconds. After this timeout, if no message
         is received, the agent executes de `idle()` method before going back
         to polling.
```
Peque: Maybe `keep_alive` and `running` attributes should be hidden too.

osbrain/nameserver.py
```diff
@@ -83,9 +83,9 @@ def __init__(self, addr=None, base=NameServer):
             addr = '127.0.0.1:%s' % addr
         self.addr = addr
         self.host, self.port = address_to_host_port(addr)
-        self.shutdown_event = multiprocessing.Event()
+        self._shutdown_event = multiprocessing.Event()
         self.uri = None
-        self.queue = multiprocessing.Queue()
+        self._queue = multiprocessing.Queue()
```
Peque: We could probably hide all attribute in the NameServerProcess as well. Couldn't we?

osbrain/agent.py
ocaballeror: Maybe `keep_alive` should be changed (it's used in one test, though).

I'm not sure about `running`, though. It's used very frequently to check if the agent is running or not. Maybe we should make the attribute private and add a method to return its value?

Peque: Yeah, an `.is_running()` method (or something like that) makes a lot of sense.

osbrain/nameserver.py
ocaballeror: I'm going to leave host, port and address public and hide the rest. They are not variables that should be changed by the user, but they might want to check their value in some cases.

osbrain/agent.py
ocaballeror: After hard deliberation I decided to only leave agent as public and make the others hidden.

osbrain/agent.py
```diff
@@ -1633,22 +1635,25 @@ def send_recv(self, address, message):
         self.send(address, message)
         return self.recv(address)
 
+    def is_running(self):
```
Peque: We should add a short docstring (i.e.: "Check whether the agent is running."). You might omit the `Returns` section if you want to, although, for consistency, it should be there too.

ocaballeror: I thought this was too self-explanatory to need a docstring, but OK 😄

osbrain/proxy.py
```diff
@@ -122,7 +122,7 @@ def wait_for_running(self, timeout=3.):
             The object itself.
         """
         time0 = time.time()
-        while not self.get_attr('running'):
+        while not self.unsafe.is_running():
```
Peque: Maybe `.is_running()` should be unsafe by default. What do you think?

If we want to check the agent is running *and* responding, then we have the  `.ping()` method.

ocaballeror: It probably should. It's a bit weird to call a method to check if the agent is running and instead get an error telling you that you can't call methods on the agent because it's not running. That is what happened to me here, by the way, and the reason why I changed this invocation to `unsafe`.

I'll change it by default 👍

ocaballeror:
@Peque I'm assuming this is what an "always unsafe" method looks like.

Peque:
@ocaballeror Probably not, as that may fail in some cases (i.e.: if you remove all `.unsafe.is_running()` in the code/tests). Or would it work?

Maybe you should simply add that method to the list in `Proxy._is_safe_method()`?

I am not sure. I can check tomorrow. :blush:

ocaballeror:
@Peque I just tried it out, and it does work, at least with the tests we have. Can you think of any scenario where it wouldn't?

But now that you mention  `._is_safe_method()`, I kinda prefer this way of doing it. It makes more sense to have the `is_running` method in the `Agent` class rather than in `Proxy`. I also tried this solution and it seems to work as well, so I'd rather go with it.

Peque:
@ocaballeror Yeah, I did not realize the method was implemented in Proxy (thought it was implemented in Agent). In that case yeah, it would work, but we do agree the `._is_safe_method` is a better approach. :blush:

PR #240, opened 2018-02-06
Add documentation on channels

### LINKED ISSUE #156: Documentation on channels
https://github.com/pytest-dev/pytest-xdist/issues/270

Peque:
Fixes #156.

### DIFF
--- a/docs/source/advanced_patterns.rst
+++ b/docs/source/advanced_patterns.rst
@@ -1,0 +1,107 @@
+.. _advanced_communication_patterns:
+
+
+*******************************
+Advanced communication patterns
+*******************************
+
+Some advanced communication patterns are also implemented in osBrain. They are
+called channels, rather than sockets, as they are formed with multiple sockets.
+
+
+.. index:: async-req-rep
+.. _async_req_rep:
+
+Asynchronous Request-Reply channel
+==================================
+
+An asynchronous request-reply channel is very similar to a normal request-reply
+pattern, only in this case we want to process the reply asynchronously (i.e.:
+assign a handler for this reply and forget about it until it is received).
+
+This means that, when using an asynchronous request-reply channel, a handler
+must be specified for both the replier and the requester.
+
+Example
+-------
+
+See the following example in which Bob sends an asynchronous request to Alice.
+Note that:
+
+- When Alice binds, she uses ``ASYNC_REP`` instead of ``REP``.
+- Bob is assigned a handler too to process Alice's reply when it is received.
+- Bob sends the request and is automatically freed. It can log messages or do
+  other stuff.
+- When the reply is received, Bob automatically processes it.
+
+.. literalinclude:: ../../examples/async_req_rep.py
+
+That is the most basic example, but channels are a bit more flexible. When
+sending an asynchronous request we may want to manually specify a handler for
+the reply. This will overwrite the default handler specified when connecting to
+the ``ASYNC_REP`` server. Try changing the ``.send()`` call::
+
+   def deaf(agent, message):
+       agent.log_info('I am deaf...')
+
+
+   bob.send('alice', 'Hello, Alice!', handler=deaf)
+
+We can also specify a maximum wait time for the reply and some code to be
+executed in case the reply was not received after that time. Try the example
+above, but setting ``wait=0.5``::
+
+   def no_reply_in_time(agent):
+       agent.log_warning('No reply received!')
+
+
+   bob.send('alice', 'Hello, Alice!', wait=0.5, on_error=no_reply_in_time)
+
+.. note:: If ``on_error`` is not specified, by default the agent will simply
+   log a warning.
+
+
+.. index:: sync-pub-sub
+.. _sync_pub_sub:
+
+Synced Publish-Subscribe channel
+================================
+
+Another common use case is the synced publish-subscribe channel. It is similar
+to the normal publish-subscribe pattern but has some extra functionality. The
+server is configured to publish messages through a ``PUB`` socket and clients
+can connect to it to receive these messages. Using the synced publish-subscribe
+channel, however, allows the clients to send requests to the publisher.
+
+This can be really useful when we want to share some data from the publisher
+with all the subscribers but we want the subscribers to be able to interact
+(and perhaps even modify) this data.
+
+.. note:: Synchronization is possible because all replies are sent to the
+   requester through the same PUB-SUB socket as the one used for normal
+   publications. Note that, even though this reply is sent through the PUB-SUB
+   socket, it will only be received by the requester and not by any other
+   subscribed client.
+
+Example
+-------
+
+In the following example we have:
+
+- A synced publisher that requires a handler for requests when binding.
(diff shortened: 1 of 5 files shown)


docs/source/advanced_patterns.rst
```diff
@@ -0,0 +1,107 @@
...
+
+Asynchronous Request-Reply channel
+==================================
+
+An asynchronous request-reply channel is very similar to a normal request-reply
+pattern, only in this case we want to process the reply asynchronously (i.e.:
+assign a handler for this reply and forget about it until it is received).
+
+This means that, when using an asynchronous request-reply channel, a handler
+must be specified not only for the replier but also for the requester.
```
ocaballeror: Maybe '... must be specified for both the replier and the requester' would sound simpler.

docs/source/advanced_patterns.rst
```diff
@@ -0,0 +1,107 @@
...
+
+Example
+-------
+
+See the following example in which Bob sends an asynchronous request to Alice.
+Note that:
+
+- When Alice binds, she uses ``ASYNC_REP`` instead of ``REP``.
+- Bob is assigned a handler too to process Alice reply when it is received.
```
ocaballeror: Alice's.

docs/source/advanced_patterns.rst
```diff
@@ -0,0 +1,107 @@
...
+- Bob sends the request and is automatically freed. It can log messages or do
+  other stuff.
+- When the reply is received, Bob automatically processes it.
+
+.. literalinclude:: ../../examples/async_req_rep.py
+
+That is the most basic example, but channels are a bit more flexible. When
+sending an asynchronous request we may want to manually specify a handler for
+the reply. This will overwrite the default handler specified when connecting to
+the ``ASYNC_REP`` server. Try to change the ``.send()`` call::
```
ocaballeror: "Try to x" sounds like the user won't be able to do so because some error will raise. Maybe change this to "Run the last example changing the .send() call", or simply "Change the .send() call".

docs/source/advanced_patterns.rst
```diff
@@ -0,0 +1,107 @@
...
+
+   def deaf(agent, message):
+       agent.log_info('I am deaf...')
+
+
+   bob.send('alice', 'Hello, Alice!', handler=deaf)
+
+We can also specify a maximum wait time for the reply and some code to be
+executed in case the reply was not received after that wait. Try the example
```
ocaballeror: Maybe change "after that wait" to "after that wait time" or simply "after that time". It sounds nicer to me 😄

docs/source/advanced_patterns.rst
```diff
@@ -0,0 +1,107 @@
...
+Synced Publish-Subscribe channel
+================================
+
+Another common use case is the synced publish-subscribe channel. It is similar
+to the normal publish-subscribe pattern but has some extra functionality. The
+server is configured to publish messages through a ``PUB`` socket and clients
+can connect to it to receive these messages. Using the synced publish-subscribe
+channel, however, allows the clients to send requests to the publisher.
+
+This can be really useful when we want to share some state/data from the
```
ocaballeror: What do you mean by "state/data"? If by "state" you mean the current status of the agent isn't that "data" after all?

Maybe simplify this to just "data" or "information" (the same goes for the sentence below).

docs/source/advanced_patterns.rst
```diff
@@ -0,0 +1,107 @@
...
+   socket, it will only be received by the requester and not by any other
+   subscribed client.
+
+Example
+-------
+
+In the following example we have:
+
+- A synced publisher that requires a handler for requests when binding.
+- Two subscribers that connect to the publisher and subscribe to all topics.
```
ocaballeror: I think you are missing a "the" in "all the topics".

docs/source/advanced_patterns.rst
```diff
@@ -0,0 +1,107 @@
...
+- The publisher starts sending publications, which are received in both
+  subscribers.
+- At some point one of the clients makes a request, which is processed by the
+  publisher. The client asynchrounously receives and processes the reply from
```
ocaballeror: There's a typo in "asynchronously"

examples/async_req_rep.py
```diff
@@ -0,0 +1,39 @@
...
+def reply_late(agent, message):
+    time.sleep(1)
+    return 'Hello, Bob!'
+
+
+def process_reply(agent, message):
+    agent.log_info('Processed reply: %s' % message)
+
+
+def deaf(agent, message):
```
ocaballeror: This function is not used in this example.

docs/source/advanced_patterns.rst
Peque: "Try" as in "experiment" or "do it yourself" (i.e.: our "prueba" not our "intenta"). Does it still sound weird to you?

docs/source/advanced_patterns.rst
Peque: It sounds better to me without the "the". And looking for search suggestions with Google...

screenshot from 2018-02-07 10-36-01

:stuck_out_tongue_winking_eye:

docs/source/advanced_patterns.rst
Peque: ausynchrounously. :joy:

examples/async_req_rep.py
Peque: Good catch! Actually... the example is wrong, it includes the changes that the user is supposed to try by themselves (i.e.: the example should not contain the wait and on-error part).

Changing it. :blush:

Peque:
@ocaballeror Fixed everything except for two things (see comments above to see if we can reach an agreement). :blush:

docs/source/advanced_patterns.rst
ocaballeror: Yeah, but in this case you are referring to all THE topics, meaning the topics you were talking about before, which is something the Google suggestions can't take into account. I'd rather add the "the", but you can leave it out if you want 😁

docs/source/advanced_patterns.rst
ocaballeror: Meh. Just a bit.

I'd also say there is a very subtle difference between "try to change" and "try changing", and I'd rather go with the latter here. But there's also no need to argue too much about this, do whatever you want 😁.

Peque: I like try changing too. :blush:

ocaballeror:
Left a couple of comments.

Basically, if you like it the way it is just say it and we can go through with the merge 👍.

docs/source/advanced_patterns.rst
Peque: As this is the first (and only) occurrence of "topics" in this documentation page, I'll leave it. :innocent: 

After all I'm saying "all topics" as in "any topic", not as "a list of topics" (i.e.: clients do not care about the topics the publisher might be using, they just want to subscribe to everything).

Peque:
@ocaballeror Liked the "changing" idea. Waiting for approval. :blush:

PR #245, opened 2018-02-08
Rename `.configure()` method to `.before_run()`

### LINKED ISSUE #243: Maybe rename `.configure()` method
@ocaballeror We can discuss this tomorrow. :wink:

Peque:
Fixes #243.

Note that I changed the documentation paragraph to avoid saying "before_run is executed when the agent is running", which can be counter-intuitive.

I was thinking we could simply end the paragraph with "Instead, you can use `before_run()`.". In example: we may not need to tell the user it is being executed right before the main loop.

@ocaballeror What do you think?

### DIFF
--- a/docs/source/introduction.rst
+++ b/docs/source/introduction.rst
@@ -55,11 +55,16 @@ saying `Hello world!` but, what exactly is happening there?
 Agents and proxies
 ==================
 
-An agent, in osBrain, is an entity that runs independly from other agents
-in the system. An agent, by default, will simply poll for incoming messages
-before executing the code defined by the developer. This means a single agent,
-as in the `Hello world!` example, makes little or no sense. Agents in a
-multi-agent system start to make sense when connected to each other.
+An agent, in osBrain, is an entity that runs independly from other agents in
+the system. When running, it executes the main loop:
+
+- Poll for incoming messages.
+- Process incoming messages and execute developer-defined code.
+- Repeat.
+
+This means a single agent, as in the `Hello world!` example, makes little or no
+sense. Agents in a multi-agent system start to make sense when connected to
+each other.
 
 The easiest way to run an agent in an osBrain architecture is by calling the
 function :func:`osbrain.run_agent <osbrain.agent.run_agent>`:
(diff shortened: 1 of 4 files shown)


ocaballeror:
@Peque we should at least mention that `.before_run()` will be called inside `.run()`. Even though most users should probably call `run_agent()` and not even worry about the `.run()` method, it's still worth mentioning when the method will be invoked. Otherwise it would look the same as `.on_init()`.

I suggest something like: 
"... instead you can use `.before_run()`, which will be invoked inside the `.run()` method."

What do you think?

Peque:
@ocaballeror I think that is perfect. Already changed that to specify it is called inside the `.run()` method.

However, now I am thinking about renaming it as `.before_loop()`... :joy:

We are saying we execute `.before_run()` *inside* `.run()`, which may confuse the user. Not saying it is wrong, just maybe a bit confusing. :thinking:

ocaballeror:
@Peque But then again, will a user know what the "loop" is? Are they supposed to understand that that is what `.run()` does (for the most part)? Do they know about the existence of such loop?

If that is something that's reasonable to expect from the users, then sure, go ahead. Otherwise, even though "`before_run()` inside  `run()`" sounds a bit weird at first, anyone can imagine how that makes sense for the implementation. At least that's how I see it 😄

Peque:
@ocaballeror I do not know, I am definitely far away from a normal user perspective, so things that seem natural to me might not be so. :joy:

Maybe your suggestion is the best. I thought that saying `.before_loop()` could teach them that there is a loop in `.run()` that the agent executes forever (no need to know about the `._loop()` method implementation, just the concept). :sweat_smile: 

Even if we leave it as `.before_run()`, do you think I should remove the last part of the paragraph? (i.e.: "and before starting the main loop"). For me it makes things clearer, but maybe it only adds noise...

Peque:
@ocaballeror Or maybe we should mention the existence of a "main loop" in "Introduction and Example", in the documentation.

ocaballeror:
@Peque  There is some mention of the loop in a few of the docstrings, so maybe it's not a bad idea to just include it in the main documentation. As someone who's more of a user of this library, and having read the documentation for the first time just a few weeks ago, I don't think it would be a bad thing to do. It's pretty easy to understand why there is a main loop and how it works, so a simple mention of it won't do any harm in my opinion.

If you decide to do it, then you can leave the note as it is, I think it's somewhat helpful to know that it's called before the main loop.

Peque:
@ocaballeror Better now? :joy:

PR #249, opened 2018-02-15
Implement two step shutdown with sigint

### LINKED ISSUE #60: Implement two step interruption
Implement two step interruption for agents/name server. First SIGINT captured results in clean exit/kill (i.e.: agent waits for code execution and tear-down and name server waits for all agents to be unregistererd). If another SIGINT is captured, kill with no mercy (as we are doing it right now).

ocaballeror:
Fixes #60 

Here's a quick rundown of the changes:

* The signal handler needs to send a shutdown signal to the agents from a separate thread, since running it in the same one as the nameserver's socket (default), would lead to a deadlock where the nameserver is waiting for the agents to shut down and the agents are waiting for the name server to accept their 'unregister' message.
* The nameserver process can't call `terminate` on itself from the sigint handler because `._popen` seems to be undefined when accessed from a different process. Right now it seems the NameserverProcess is being terminated anyway (the tests would block otherwise, and there's no zombie python process afterwards), but I'm not entirely sure how or why.
* I removed the sigint handler on the AgentProcess. I don't think it's needed anymore, since we are capturing the signal from the nameserver.
* This took me way longer than I expected, so I hope this is the right way to do it 👍


ocaballeror:
More changes:
* Re-added and fixed the sigint handler for agents.
* Added a test for the agent's sigint handler based on the commits from this branch: https://github.com/Flood1993/osbrain/tree/sigint_tests
* Added psutil as a testing dependency. It makes process management much easier to handle, especially if you are dealing with multiple OSs.

@Peque Let me know how much you disagree with me 😄

ocaballeror:
@Peque Is this a good moment to go on with this?

This should be rebased on top of the appveyor integration, by the way, I don't trust Windows signals whatsoever.

Peque:
@ocaballeror Let us first improve the current Windows CI integration (combine coverage reports, fix tests currently failing, see if we can skip less tests...).

Trying to tackle this now would only add more noise to the currently incomplete Windows support.

ocaballeror:
@Peque Shall we take a look at this again?

Peque:
@ocaballeror Sure. I will review this this week.

Peque:
@ocaballeror Before reviewing this, would you mind rebasing against `master`?

PR #251, opened 2018-02-15
Safer one-way kill in close IPC socket test

Peque:
Hopefully prevents this from happening again:

https://travis-ci.org/opensistemas-hub/osbrain/jobs/341672571

### DIFF
--- a/osbrain/helper.py
+++ b/osbrain/helper.py
@@ -195,6 +195,36 @@ def wait_agent_attr(agent, name='received', length=None, data=None, value=None,
     return False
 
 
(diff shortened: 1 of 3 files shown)


Peque:
@ocaballeror Yeah, I bet there are a couple of tests that could benefit from this. I'll implement the "negated wait_condition()" too in that case and ask you for review again. We can then open a ticket to have a look at where we could use this function.

Peque:
@ocaballeror Added a `negate` parameter, which should be useful if we integrate the use of this function in other tests in the near future.

PR #252, opened 2018-02-20
Make tests compatible with windows

ocaballeror:
Continuing the discussion from #147 .

Also, just to be clear, I don't intend to merge this yet, so there may be some weird comments and print statements around the code.

### DIFF
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -1817,6 +1817,7 @@ class AgentProcess(multiprocessing.Process):
         try:
             ns = NSProxy(self.nsaddr)
             self._daemon = Pyro4.Daemon(self._host, self.port)
+            self.base = cloudpickle.loads(self.base)
             self.agent = self.base(name=self.name, host=self._host,
                                    serializer=self._serializer,
                                    transport=self._transport,
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -298,7 +298,7 @@ class Agent():
         """
         Handle incoming messages in the loopback socket.
         """
-        header, data = dill.loads(message)
+        header, data = cloudpickle.loads(message)
         if header == 'EXECUTE_METHOD':
             method, args, kwargs = data
             try:
@@ -316,7 +316,7 @@ class Agent():
         """
         Handle incoming messages in the _loopback_safe socket.
         """
-        method, args, kwargs = dill.loads(data)
+        method, args, kwargs = cloudpickle.loads(data)
         try:
             response = getattr(self, method)(*args, **kwargs)
         except Exception as error:
@@ -345,7 +345,7 @@ class Agent():
         """
         if not self._running:
             raise NotImplementedError()
-        data = dill.dumps((header, data))
+        data = cloudpickle.dumps((header, data))
         return self._loopback_reqrep('inproc://loopback', data)
 
     def safe_call(self, method, *args, **kwargs):
@@ -366,7 +366,7 @@ class Agent():
         if not self._running:
             raise RuntimeError(
                 'Agent must be running to safely execute methods!')
-        data = dill.dumps((method, args, kwargs))
+        data = cloudpickle.dumps((method, args, kwargs))
         return self._loopback_reqrep('inproc://_loopback_safe', data)
 
     def each(self, period, method, *args, alias=None, **kwargs):
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -1801,7 +1801,7 @@ class AgentProcess(multiprocessing.Process):
         self.nsaddr = nsaddr
         self._serializer = serializer
         self._transport = transport
-        self.base = base
+        self.base = cloudpickle.dumps(base)
         self._shutdown_event = multiprocessing.Event()
         self._queue = multiprocessing.Queue()
         self._sigint = False
--- a/osbrain/tests/common.py
+++ b/osbrain/tests/common.py
@@ -1,9 +1,23 @@
+import sys
 import pytest
+from pytest import mark
 
 from osbrain import run_agent
 from osbrain import run_logger
 from osbrain import run_nameserver
 from osbrain.helper import sync_agent_logger
+
+skip_windows = mark.skipif(sys.platform == 'win32',
+                           reason='Not supported on windows')
+skip_windows_port_reuse = mark.skipif(sys.platform == 'win32',
+                                      reason='Windows allows port reuse')
+skip_windows_any_port = mark.skipif(sys.platform == 'win32',
+                                    reason='Windows allows binding to well '
+                                    'known ports')
+skip_windows_spawn = mark.skipif(sys.platform == 'win32',
+                                 reason='Windows does not support fork')
+skip_windows_ipc = mark.skipif(sys.platform == 'win32',
+                               reason='Windows does not support IPC')
 
 
 def append_received(agent, message, topic=None):
--- a/osbrain/tests/test_agent.py
+++ b/osbrain/tests/test_agent.py
@@ -28,6 +28,10 @@ from common import nsproxy  # pragma: no flakes
 from common import append_received
 from common import set_received
 
+from common import skip_windows_spawn
+from common import skip_windows_any_port
+from common import skip_windows_port_reuse
+
 
 def test_agent_uuid():
     """
@@ -265,6 +269,7 @@ def test_bind_tcp_addr_specific_port(nsproxy):
     assert address.address.port == port
 
 
+@skip_windows_spawn
 @pytest.mark.parametrize('linger, sleep_time, should_receive', [
     (2, 1, True),
     (0.5, 1, False),
@@ -599,6 +604,7 @@ def test_running_exception(nsproxy):
     assert not agent.is_running()
 
 
+@skip_windows_port_reuse
 def test_agent_error_address_already_in_use(nsproxy):
     """
     Running an agent should raise an error if address is already in use.
@@ -609,6 +615,7 @@ def test_agent_error_address_already_in_use(nsproxy):
     assert 'Address already in use' in str(error.value)
 
 
+@skip_windows_any_port
 def test_agent_error_permission_denied(nsproxy):
     """
     Running an agent should raise an error if it has not sufficient
--- a/osbrain/tests/test_agent.py
+++ b/osbrain/tests/test_agent.py
@@ -37,7 +37,7 @@ def test_agent_uuid():
     """
     All agent identifiers should be unique strings.
     """
-    N = 1000
+    N = 100
     bunch = set(Agent()._uuid for i in range(N))
     assert len(bunch) == N
     assert all(isinstance(identifier, bytes) for identifier in bunch)
--- a/osbrain/tests/test_agent.py
+++ b/osbrain/tests/test_agent.py
@@ -306,7 +306,7 @@ def test_linger(nsproxy, linger, sleep_time, should_receive):
     puller.bind('PULL', alias='pull', handler=append_received,
                 addr=address.address, transport='tcp')
 
-    assert should_receive == wait_agent_attr(puller, data='foo', timeout=.2)
+    assert should_receive == wait_agent_attr(puller, data='foo', timeout=1)
 
 
 def test_pushpull(nsproxy):
--- a/osbrain/tests/test_agent_ipc_sockets.py
+++ b/osbrain/tests/test_agent_ipc_sockets.py
@@ -9,6 +9,9 @@ from osbrain.helper import agent_dies
 from osbrain.helper import wait_condition
 
 from common import nsproxy  # pragma: no flakes
+from common import skip_windows_ipc
+
+pytestmark = skip_windows_ipc
 
 
 def test_agent_close_ipc_socket_agent_shutdown(nsproxy):
--- a/osbrain/tests/test_agent_transport.py
+++ b/osbrain/tests/test_agent_transport.py
@@ -15,27 +15,37 @@ from osbrain.helper import wait_agent_attr
 
 from common import nsproxy  # pragma: no flakes
 from common import append_received
+from common import skip_windows_spawn
+from common import skip_windows_ipc
 
 
-def test_agent_bind_transport_global(nsproxy):
+def test_agent_bind_transport_platform_default(nsproxy):
     """
-    Test global default transport.
+    Default transport is platform-dependent.
     """
-    # Default transport
     agent = run_agent('a0')
     address = agent.bind('PUSH')
-    assert address.transport == 'ipc'
+    if os.name == 'posix':
+        assert address.transport == 'ipc'
+    else:
+        assert address.transport == 'tcp'
 
-    # Changing default global transport
-    osbrain.config['TRANSPORT'] = 'tcp'
+
+@skip_windows_spawn
+def test_agent_bind_transport_global(nsproxy):
+    """
+    Test global default transport change.
+    """
+    # Default transport is not `inproc`
+    agent = run_agent('a0')
+    address = agent.bind('PUSH')
+    assert address.transport != 'inproc'
+
+    # Changing default global transport to `inproc`
+    osbrain.config['TRANSPORT'] = 'inproc'
     agent = run_agent('a1')
     address = agent.bind('PUSH')
-    assert address.transport == 'tcp'
-
-    osbrain.config['TRANSPORT'] = 'ipc'
-    agent = run_agent('a2')
-    address = agent.bind('PUSH')
-    assert address.transport == 'ipc'
+    assert address.transport == 'inproc'
 
 
 def test_agent_bind_transport_agent(nsproxy):
@@ -46,9 +56,9 @@ def test_agent_bind_transport_agent(nsproxy):
     address = agent.bind('PUSH')
     assert address.transport == 'tcp'
 
-    agent = run_agent('a1', transport='ipc')
+    agent = run_agent('a1', transport='inproc')
     address = agent.bind('PUSH')
-    assert address.transport == 'ipc'
+    assert address.transport == 'inproc'
 
 
 def test_agent_bind_transport_bind(nsproxy):
@@ -64,18 +74,21 @@ def test_agent_bind_transport_bind(nsproxy):
     assert address.transport == 'inproc'
 
 
-def test_agent_bind_given_address(nsproxy):
+@skip_windows_ipc
+def test_agent_bind_given_address_ipc(nsproxy):
     """
     Test agent binding to an specified address using TCP and IPC transport
     layers.
     """
     agent = run_agent('a0')
-    # IPC
     ipc_addr = str(uuid4())
     address = agent.bind('PUSH', addr=ipc_addr, transport='ipc')
     assert address.transport == 'ipc'
     assert address.address.name == ipc_addr
-    # TCP
+
+
+def test_agent_bind_given_address_tcp(nsproxy):
+    agent = run_agent('a0')
     while True:
         try:
             # Bind to random port
@@ -90,6 +103,7 @@ def test_agent_bind_given_address(nsproxy):
     assert address.address == tcp_addr
 
 
+@skip_windows_ipc
 def test_agent_ipc_from_different_folders(nsproxy):
     """
     IPC should work well even when agents are run from different folders.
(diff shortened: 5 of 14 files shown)


ocaballeror:
Also, it turns out many tests were failing due to timeouts because we were trying to run 8 parallel threads when the appveyor machine has only 2 cores. Using `pytest -n2` works much better :+1:

Peque:
@ocaballeror Yeah, we could even avoid using `pytest-xdist` in AppVeyor. Historically it has given some problems in Travis, but at least it sped things up. If we are not going to notice the speed-up much, then we might as well avoid using it.

I'd say we could even, for AppVeyor, run tests in 1 thread (i.e.: no `pytest-xdist`) and only against the latest Python version.

ocaballeror:
@Peque I've run it a few times with 2 threads and it seems to work fine. We could leave it with 2 and revert to single-threaded mode if we run into any problems in the future. After all, if we have extra cores, why not use them?

About the  python versions, though, why test only the latest one? There could be some implementation differences from version to version that suddenly break compatibility, so I'd leave it as it is right now just to be safe.

ocaballeror:
Update: Still working on the linger tests. There must be something that Windows is doing differently in the internal implementation of sockets and the SO_LINGER option, but I'm having a hard time figuring out what it is.

I'll keep updating when I find more stuff.

ocaballeror:
Took me a long while to figure it out. In the end it turns out the problem with the linger tests had nothing to do with sockets, zmq or anything related to that. The problem was actually being caused by the way we changed the linger value in every test. We were updating the global `config` dictionary, and then expecting the `.close_all()` method to read the updated value, but that won't work on Windows because of the way it creates new processes. On Unix the `fork()` system called is used to create the agent processes, but since that is not available on Windows, and the new processes are created using `spawn()`, we can't expect the global variable `config` to be copied to child processes.

In the end, the call to `get_linger()` inside `close_all()` was returning the default value of 1000 for `linger` instead of the updated one. I changed this so `,close_all()` accepts a `linger` parameter.

ocaballeror:
By the way, there are 2 other tests currently failing and they are both related to Windows allowing more than one socket to listen on one address.

I will try to implement some kind of check to stop this from happening and manually raise an error if the user tries to do that. @Peque or should I just skip these tests on Windows and kindly advise people against this sort of practice?

osbrain/tests/test_agent_transport.py
```diff
@@ -24,18 +25,19 @@ def test_agent_bind_transport_global(nsproxy):
...
 
-    osbrain.config['TRANSPORT'] = 'ipc'
-    agent = run_agent('a2')
-    address = agent.bind('PUSH')
-    assert address.transport == 'ipc'
+    if os.name == 'posix':
+        osbrain.config['TRANSPORT'] = 'ipc'
+        agent = run_agent('a2')
+        address = agent.bind('PUSH')
+        assert address.transport == 'ipc'
```
Peque: I think there should probably be a test for global transport configuration, in example:

```python
def test_agent_bind_transport_platform_default(nsproxy):
    """
    Default transport is platform-dependent.
    """
    agent = run_agent('a0')
    address = agent.bind('PUSH')
    if os.name == 'posix':
        assert address.transport == 'ipc'
    else:
        assert address.transport == 'tcp'
```

And then another for the global setting:

```python
def test_agent_bind_transport_global(nsproxy):
    """
    Test global default transport change.
    """
    # Default transport is not `inproc`
    agent = run_agent('a0')
    address = agent.bind('PUSH')
    assert address.transport != 'inproc'

    # Changing default global transport to `inproc`
    osbrain.config['TRANSPORT'] = 'inproc'
    agent = run_agent('a1')
    address = agent.bind('PUSH')
    assert address.transport == 'inproc'
```

osbrain/tests/test_agent.py
```diff
@@ -292,7 +294,7 @@ def on_init(self):
     assert agent_dies('puller', nsproxy)
 
     pusher.send('push', 'foo')
-    pusher.close_all()
+    pusher.close_all(get_linger())
```
Peque: This should be removed (see comments on linger changes).

osbrain/tests/test_agent_transport.py
```diff
@@ -46,9 +48,10 @@ def test_agent_bind_transport_agent(nsproxy):
     address = agent.bind('PUSH')
     assert address.transport == 'tcp'
 
-    agent = run_agent('a1', transport='ipc')
-    address = agent.bind('PUSH')
-    assert address.transport == 'ipc'
+    if os.name == 'posix':
+        agent = run_agent('a1', transport='ipc')
+        address = agent.bind('PUSH')
+        assert address.transport == 'ipc'
```
Peque: Just use `inproc` instead of `ipc` to make it work in all platforms.

osbrain/agent.py
```diff
@@ -1759,15 +1759,15 @@ def close(self, alias):
...
+    def close_all(self, linger=get_linger()):
         """
         Close all non-internal zmq sockets.
         """
         # Each socket might be pointed by different keys
         sockets_to_delete = []
         for sock in self._get_unique_external_zmq_sockets():
             sockets_to_delete.append(sock)
-            sock.close(linger=get_linger())
+            sock.close(linger)
```
Peque: This is not the expected behavior.

Do you know when that `get_linger()` gets executed?

In my opinion we should leave it as it was before and skip if Windows fails, then create a new issue to investigate this issue and/or a workaround (i.e.: serializing the global configuration to child processes?). But for now it is easier if we simply skip.

osbrain/tests/test_proxy.py
```diff
@@ -57,7 +57,7 @@ def test_wrong_nameserver_address(timeout):
     t0 = time.time()
     with pytest.raises(TimeoutError):
         locate_ns('127.0.0.1:1234', timeout=timeout)
-    assert timeout <= time.time() - t0 <= timeout + 1.
+    assert timeout <= time.time() - t0 <= timeout + 2.
```
Peque: Was this failing with the previous timeout?

osbrain/tests/test_proxy.py
```diff
@@ -323,4 +323,4 @@ def shoot(self):
     assert time.time() - t0 < 0.2
     assert not wayne._next_oneway
 
-    assert wait_agent_attr(wayne, value=20*['bang!'], timeout=1.2)
+    assert wait_agent_attr(wayne, value=20*['bang!'], timeout=1.5)
```
Peque: Was this failing with the previous timeout?

osbrain/tests/test_nameserver.py
```diff
@@ -342,8 +344,10 @@ def test_random_nameserver_process():
     # Raising exceptions
     with pytest.raises(ValueError):
         random_nameserver_process(port_start=-1, port_stop=-2)
-    with pytest.raises(RuntimeError):
-        random_nameserver_process(port_start=22, port_stop=22, timeout=0.5)
+    if sys.platform != 'win32':
+        with pytest.raises(RuntimeError):
+            random_nameserver_process(port_start=22,
+                                      port_stop=22, timeout=0.5)
```
Peque: I'd say we simply skip this test on Windows for now and revert the changes.

osbrain/tests/test_agent_transport.py
```diff
@@ -71,10 +74,12 @@ def test_agent_bind_given_address(nsproxy):
...
-    ipc_addr = str(uuid4())
-    address = agent.bind('PUSH', addr=ipc_addr, transport='ipc')
-    assert address.transport == 'ipc'
-    assert address.address.name == ipc_addr
+    if os.name == 'posix':
+        ipc_addr = str(uuid4())
+        address = agent.bind('PUSH', addr=ipc_addr, transport='ipc')
+        assert address.transport == 'ipc'
+        assert address.address.name == ipc_addr
+
```
Peque: I'd say we split this test in two: one for IPC, one for TCP. Add then a skip-on-windows mark for the IPC test.

Peque:
@ocaballeror I would say, for now, we just skip failing tests for Windows. Then we can create specific issues for those fails and decide whether we should fix them or just assume they should fail in Windows.

Also, that way we keep this PR a bit simpler and merge it faster (even if it is not perfect, it is definitely a step forward that we want to include in our code base). :blush:

osbrain/tests/test_agent.py
```diff
@@ -23,17 +23,19 @@
...
 from common import set_received
+from common import skip_on_windows
 
 
 def test_agent_uuid():
     """
     All agent identifiers should be unique strings.
     """
-    N = 1000
+    N = 100
```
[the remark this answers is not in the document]
ocaballeror: Yes. I don't understand why, but for some reason, it takes 12-13 seconds to complete on Windows and about 0.2 seconds on Linux. I thought `N=100` would be enough for this test anyway

osbrain/tests/test_agent_transport.py
ocaballeror: That reminds me that changing values in the global config does not work on windows, for the same reason I explained earlier when I was talking about the linger thing.
When calling `run_agent()` a new process will be created, but because of the way Windows creates child processes, the values in `config` won't be the ones we just set, but rather the default ones.

I'll add these two tests and try to find a way to make the second one work on Windows.

osbrain/tests/test_agent.py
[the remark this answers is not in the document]
Peque: I knew Windows processes were slow, but did not know they were *that* slow. The main difference is the way they are created (spawn vs. fork). Forking is definitely much faster, specially with the copy-on-write implemented in Linux.

Leave it with 100 then. :+1:

osbrain/tests/test_agent_transport.py
ocaballeror: Oh. I was convinced 'IPC' and 'inproc' were the same thing. Good thing you pointed it out.

osbrain/tests/test_agent_transport.py
Peque: As I mention bellow, I'd leave that until we merge this PR. Just having AppVeyor integrated, even if we are skipping some tests in Windows, is worth it to merge. Then we can see how to gradually reduce the number of skipped tests.

osbrain/tests/test_proxy.py
ocaballeror: At least on my machine, yes. :smile:

osbrain/tests/test_proxy.py
ocaballeror: Maybe at some point, I wouldn't have changed it otherwise. I'll run it through appveyor again, maybe it needed longer timeouts before, since we were using 8 threads instead of 2.

osbrain/tests/test_nameserver.py
ocaballeror: Makes sense.

osbrain/tests/test_agent_transport.py
Peque: `inproc` is for in-process communication (i.e.: between threads). Happens in memory, so it is much faster than IPC.

osbrain/tests/test_nameserver.py
ocaballeror: What about `random_nameserver_process()` then? Shouldn't we test it works?

osbrain/agent.py
ocaballeror: Why is it not the expected behavior? I don't understand what you mean. The value of linger can be passed as an optional parameter, and if it isn't, `get_linger()` is called, as it was before. How does this differ from the expected behavior?

osbrain/tests/test_proxy.py
ocaballeror: Turns out 1.2 is enough, but with just 1, the test will fail everytime, both on my machine and appveyor

ocaballeror:
@Peque Maybe a should create another flag? Something like `skip_for_now` so we know that specific test still needs reviewing?

osbrain/agent.py
```diff
@@ -1816,7 +1816,10 @@ def run(self):
 
         try:
             ns = NSProxy(self.nsaddr)
+            reuse = Pyro4.config.SOCK_REUSE
+            Pyro4.config.SOCK_REUSE = False
             self._daemon = Pyro4.Daemon(self._host, self.port)
+            Pyro4.config.SOCK_REUSE = reuse
```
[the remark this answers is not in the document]
ocaballeror: It defaults to `True`. And for some reason (I haven't investigated much), everything fails spectacularly if the default value is changed to `False` in `__init__.py`.

But OK, I'll revert these two changes and skip the tests that rely on OS errors being raised from binding to addresses in use.

osbrain/tests/test_nameserver.py
Peque: For now let us just skip those failing tests. We will fix them later, as separate PRs, analyzing the situation for each case.

Peque:
@ocaballeror I fear the `skip_for_now` name... :joy: (you know, sometimes "temporary" fixes end up being permanent...)

I like the idea of using more specific flags though. Maybe `skip_windows_ipc` and a generic `skip_windows_broken` for (currently) unknown problems.

osbrain/agent.py
Peque:
```python
conf = dict(linger=1)


def get_linger():
    return conf['linger']


def default_conf(linger=get_linger()):
    return linger


if __name__ == '__main__':
    print(default_conf())
    conf['linger'] = 2
    print(default_conf())
```

osbrain/agent.py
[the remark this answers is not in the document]
ocaballeror: Sure

osbrain/agent.py
ocaballeror: Oh. I didn't know that happened. What about something like this:
```python
conf = dict(linger=1)


def get_linger():
    return conf['linger']


def default_conf(linger=None):
    if linger is None:
        linger = get_linger()
    return linger


if __name__ == '__main__':
    print(default_conf())
    conf['linger'] = 2
    print(default_conf())
```

Peque: Yeah, although that should be in a separate PR. Basically you'd be adding support for passing a linger parameter to .close_all(). That would require new tests. If you want to implement this we should also probably add the same parameter to .close() (with tests again).

As it is a non-trivial change, I would not add it here. Let us for now keep this PR for only making AppVeyor succeed (even if skipping some tests for now).

ocaballeror: Ok. I will skip the linger tests as well, then.

osbrain/tests/common.py
```diff
@@ -1,10 +1,16 @@
...
 
 from osbrain import run_agent
 from osbrain import run_logger
 from osbrain import run_nameserver
 from osbrain.helper import sync_agent_logger
 
+skip_on_windows = pytest.mark.skipif(sys.platform == 'win32',
+                                     reason='Not supported on windows')
+skip_windows_addr = pytest.mark.skipif(sys.platform == 'win32',
+                                       reason='Windows allows port reuse')
```
Peque: Just for consistency I'd name both `skip_on_windows...` or `skip_windows...` (it is easier to `grep`).

osbrain/tests/test_agent.py
```diff
@@ -599,16 +603,18 @@ def test_running_exception(nsproxy):
...
     Running an agent should raise an error if address is already in use.
     """
     with pytest.raises(RuntimeError) as error:
         run_agent('a0', nsaddr=nsproxy.addr(), addr=nsproxy.addr())
     assert 'OSError' in str(error.value)
-    assert 'Address already in use' in str(error.value)
 
 
+@pytest.mark.skipif(sys.platform == 'win32',
+                    reason='Windows allows binding to any port')
```
Peque: Even if only used once, I'd declare all Windows skips the same way (with a `@skip...` decorator). Will make finding all Windows skips easier.

osbrain/tests/test_agent_transport.py
Peque: You forgot to change this (i.e.: create 2 separate tests). :wink:

osbrain/tests/test_agent_transport.py
Peque: I still think we should split this in 2.

osbrain/tests/test_agent_transport.py
```diff
@@ -90,6 +94,7 @@ def test_agent_bind_given_address(nsproxy):
     assert address.address == tcp_addr
 
 
+@skip_on_windows
```
Peque: We could use `skip_on_windows_ipc` to be more clear.

osbrain/tests/test_nameserver.py
```diff
@@ -346,16 +349,18 @@ def test_random_nameserver_process():
...
     Name server start() should raise an error if address is already in use.
     """
     with pytest.raises(RuntimeError) as error:
         run_nameserver(nsproxy.addr())
     assert 'OSError' in str(error.value)
-    assert 'Address already in use' in str(error.value)
 
 
+@pytest.mark.skipif(sys.platform == 'win32',
+                    reason='Windows allows binding to any port')
```
Peque: Rather use `@skip...` everywhere.

osbrain/tests/test_agent_ipc_sockets.py
```diff
@@ -9,6 +10,9 @@
 
 from common import nsproxy  # pragma: no flakes
 
+pytestmark = pytest.mark.skipif(sys.platform == 'win32',
+                                reason='IPC not available on Windows')
```
Peque: Does `pytestmark = skip_windows_ipc` work? (importing the corresponding `skip_windows_ipc` definition).

This way we put all `pytest.mark.skipif` declarations together.

osbrain/tests/test_nameserver.py
```diff
@@ -346,16 +349,17 @@ def test_random_nameserver_process():
...
 
+@skip_windows_port_reuse
 def test_nameserver_oserror(nsproxy):
     """
     Name server start() should raise an error if address is already in use.
     """
     with pytest.raises(RuntimeError) as error:
         run_nameserver(nsproxy.addr())
     assert 'OSError' in str(error.value)
-    assert 'Address already in use' in str(error.value)
```
Peque: You removed this line but now that we are skipping for Windows, it should be there.

osbrain/tests/test_agent.py
```diff
@@ -301,7 +306,7 @@ def on_init(self):
     puller.bind('PULL', alias='pull', handler=append_received,
                 addr=address.address, transport='tcp')
 
-    assert should_receive == wait_agent_attr(puller, data='foo', timeout=.2)
+    assert should_receive == wait_agent_attr(puller, data='foo', timeout=1)
```
Peque: I guess this timeout does not need to be changed for now as this test is skipped for Windows. When/if we remove the skip we will change the timeout if necessary.

osbrain/tests/test_agent.py
```diff
@@ -599,16 +604,17 @@ def test_running_exception(nsproxy):
...
 
+@skip_windows_port_reuse
 def test_agent_error_address_already_in_use(nsproxy):
     """
     Running an agent should raise an error if address is already in use.
     """
     with pytest.raises(RuntimeError) as error:
         run_agent('a0', nsaddr=nsproxy.addr(), addr=nsproxy.addr())
     assert 'OSError' in str(error.value)
-    assert 'Address already in use' in str(error.value)
```
Peque: You removed this line but now that we are skipping for Windows, it should be there.

osbrain/tests/test_proxy.py
Peque: Is still fails if set back to `1.2`?

osbrain/tests/test_agent.py
ocaballeror: But this one does fail on Linux from time to time. I ran it in a loop yesterday, and it took me anywhere between 50 and 500 attempts to make it fail, but eventually it does.

It has failed a few times in travis:
https://travis-ci.org/ocaballeror/osbrain/jobs/346269407

I think 1 is a safer value.

Peque: @ocaballeror Then ok. It seems changing the timeout does not actually change the test, will only make it a bit slower for the cases where `should_receive == False`, so it is fine for me. :blush:

Should we put it as a separate commit? i.e.: "More lenient timeout in linger test". Simply because it is not really related to Windows tests.

ocaballeror: Makes sense 👍 

Rebasing this is going to be fun....

osbrain/tests/test_proxy.py
ocaballeror: Same as above, I ran it in a loop to see if it would eventually fail, and after 100-200 tests I've gotten it to fail a couple of times. 1.5 looks like a safer value.

ocaballeror:
Finally 😄 

This took way longer than I expected

PR #254, opened 2018-02-26
Use `wait_condition` in agent IPC tests

Peque:
Hopefully prevents this from happening again:

https://travis-ci.org/opensistemas-hub/osbrain/jobs/346389474


ocaballeror:
@Peque What about the other three tests? They could probably benefit from using `wait_condition()` instead of `time.sleep()` + `assert not address.exists()`

Peque:
@ocaballeror In the other 3 tests there are no crashes/killings (just clean shutdowns), so it is very unlikely that the file is still there by the time we check, but I am not 100% sure it is impossible, so I have changed those too, just in case. :+1:

PR #257, opened 2018-02-27
Update docs with default serializer change

Peque:


docs/source/security.rst
```diff
@@ -18,13 +18,16 @@ Security
...
-osBrain uses :py:mod:`pickle` module for serialization when passing messages
-between agents and can use :py:mod:`pickle` and :py:mod:`dill` for
-serialization when configuring and deploying the multi-agent architectures.
-It is well known that using pickle or dill for this purpose is a security risk.
-The main problem is that allowing a program to unpickle or undill arbitrary
-data can cause arbitrary code execution and this may wreck or compromise your
-system.
+osBrain can use different serializers when passing messages between agents and
+when configuring and deploying the multi-agent architectures. Among this
+serializers, some are considered unsafe (:py:mod:`pickle`,
```
ocaballeror: Only "some" are unsafe? Is there any safe serializer?

docs/source/security.rst
```diff
@@ -18,13 +18,16 @@ Security
...
+:py:mod:`cloudpickle`...).
+
+Using these unsafe serializers is a security risk. The main problem is that
+allowing a program to deserialize arbitrary data can cause arbitrary code
+execution and this may wreck or compromise your system. Therefore, osBrain is
+meant to be run only within trusted networks (i.e.: LANs) or with properly
+encrypted/safe communications (see `Protocol encryption`_).
```
ocaballeror: Is the space after Protocol encryption necessary?

ocaballeror: Also, `serialization.rst` states that we use `pickle` by default. Is that right?

ocaballeror:
@Peque Also, `serialization.rst` states that we use pickle by default. Is that right?

docs/source/security.rst
Peque: Yeah, JSON, raw... Those can be used for communication between agents.

docs/source/security.rst
Peque: Yeah, it creates a link to that section in the documentation.

Peque:
@ocaballeror That is correct. Currently we use `pickle` by default for communication between agents. Not the most capable, but it is the fastest (and usually between agents you pass pickable data).

PR #260, opened 2018-02-28
More robust linger test implementation

### LINKED ISSUE #250: test_linger[2-1-True]
- https://travis-ci.org/opensistemas-hub/osbrain/jobs/341786321
- https://travis-ci.org/opensistemas-hub/osbrain/jobs/341672572
- https://travis-ci.org/Peque/osbrain/jobs/341815179
- https://travis-ci.org/Peque/osbrain/jobs/342288465

Peque:
Trying two things:

- Make sure the connection is well established before shutting down the pusher (i.e.: actually send a message between the agents).
- Put the puller second start before the sleep to avoid adding more seconds to that sleep if the starting takes too long.

Hopefully fixes #250.

### DIFF
--- a/osbrain/tests/test_agent.py
+++ b/osbrain/tests/test_agent.py
@@ -293,16 +293,25 @@ def test_linger(nsproxy, linger, sleep_time, should_receive):
                           transport='tcp')
 
     pusher.connect(address, alias='push')
+
+    # Make sure connection is well established
+    pusher.send('push', 'ping')
+    assert wait_agent_attr(puller, data='ping', timeout=1)
+
+    # Shutdown the puller and restart it without binding
     puller.shutdown()
     assert agent_dies('puller', nsproxy)
+    puller = run_agent('puller', base=AgentTest)
 
+    # Push a new message, which should block during linger period
     pusher.send('push', 'foo')
     pusher.close_all()
+
     # After this timeout, depending on linger value, 'foo' will no longer be
     # on queue to be sent
     time.sleep(sleep_time)
 
-    puller = run_agent('puller', base=AgentTest)
+    # Bind to receive the message (if still in queue)
     puller.bind('PULL', alias='pull', handler=append_received,
                 addr=address.address, transport='tcp')
 


ocaballeror:
I want to test it thoroughly first.

For some reason, I can only get the builds to fail on travis, and  not on any of my machines. Even the least powerful of them (a single core VM with 100%  CPU usage) managed to run the linger test 800 times without a single failure.

I just scheduled a few builds with your commit on Travis, so let's see how that goes.
https://travis-ci.org/ocaballeror/osbrain/builds

Peque:
@ocaballeror It seems all linger tests passed. Other errors appeared though...

- [test_agent_close_ipc_socket_agent_blocked_nameserver_shutdown](https://travis-ci.org/ocaballeror/osbrain/jobs/347274145)
- [test_agent_close_ipc_socket_agent_kill](https://travis-ci.org/ocaballeror/osbrain/jobs/347261263)
- [???](https://travis-ci.org/ocaballeror/osbrain/jobs/347264537)

About the latter, it is impossible for us to know what happened. For the first 2 we could open an issue and try to reproduce the same error (although you know reproducing those errors locally is sometimes hard...). :sweat:

Anyway this PR probably fixes the lingering problem. Feel free to do more tests if you want. Otherwise we can merge it and focus on the current Windows problems or these socket file test problems.

ocaballeror:
Yeah, I ran the tests 20 times and some of them failed, but it was never the linger one, so it's probably safe to assume this fixes it.

About the other failing tests, though, I just realized I ran all of them at the same time, meaning I had 10 builds running concurrently, each of them testing 3 different versions of python. It just came to my mind that Travis is probably not creating 10 VMs just for us, so all of the builds were running on the same machine. I will have to re run them one by one and see if there are still other tests failing.
Regardless of this, I wrote down which tests were failing, why and where, so if we find out they are legitimally failing (and not just because of the resource overloading I was talking about before), we at least have a starting point to conduct an investigation.

Let me try this, and I will get back to you with the results (if I ever find time to do it :smile: )

Peque:
@ocaballeror is this okay to be merged?

Issue #259, opened 2018-02-27
Windows tests seem to fail more often


Peque:
Test | Line
--- | ---
[`test_agent_loopback_header_unknown`](https://ci.appveyor.com/project/osbrain-ci/osbrain/build/job/63au4xrqvxni08jk) | osbrain\tests\common.py:35:
[`test_async_rep_send_handler_types[<lambda>-False]`](https://ci.appveyor.com/project/osbrain-ci/osbrain/build/1.0.8/job/93pvxghvv9945qpp) | osbrain\tests\common.py:35:
[`test_log_levels`](https://ci.appveyor.com/project/osbrain-ci/osbrain/build/1.0.9/job/o87p70i27n2uwknk) | osbrain\tests\common.py:35:
[`test_nameserver_proxy_shutdown_lazy_agents[True-5]`](https://ci.appveyor.com/project/osbrain-ci/osbrain/build/1.0.9/job/o87p70i27n2uwknk) | osbrain\tests\test_nameserver.py:155:
[`test_pubsub_topics_raw[raw]`](https://ci.appveyor.com/project/osbrain-ci/osbrain/build/1.0.9/job/byg0wi8cutaoe3v4) | osbrain\tests\common.py:35:
[`test_pubsub_topics_separator[cloudpickle]`](https://ci.appveyor.com/project/ocaballeror/osbrain-5qldx/build/1.0.1/job/xusdm84lmqsw19yl) | osbrain\tests\common.py:35:
[`test_pubsub_topics_separator[cloudpickle]`](https://ci.appveyor.com/project/osbrain-ci/osbrain/build/1.0.8/job/93pvxghvv9945qpp) | osbrain\tests\common.py:35:
[`test_pubsub_topics_separator[json]`](https://ci.appveyor.com/project/osbrain-ci/osbrain/build/1.0.9/job/o87p70i27n2uwknk) | osbrain\tests\common.py:35:
[`test_pubsub_topics_separator[pickle]`](https://ci.appveyor.com/project/osbrain-ci/osbrain/build/1.0.8/job/05t52l0fccmxs9mg) | osbrain\tests\common.py:35:
[`test_pushpull[json-Hello world]`](https://ci.appveyor.com/project/osbrain-ci/osbrain/build/1.0.3/job/fqbkoglohcpmi5bl) | osbrain\tests\common.py:35:
[...]

PR #262, opened 2018-03-05
More robust stop-timer tests implementation

Peque:
These tests were failing from time to time. Now they are more robust and fast (avoiding unnecessary time sleeps).

### DIFF
--- a/osbrain/tests/test_helper.py
+++ b/osbrain/tests/test_helper.py
@@ -118,25 +118,44 @@ def test_logger_received_count(agent_logger):
     assert logger_received(logger, message='foo.*bar') == 2
 
 
-@pytest.mark.parametrize('attribute,length,data,value,match', [
-    ([], 1, None, None, False),
-    ([], None, 1, None, False),
-    ([], None, None, [], True),
-    ([1, 2], 2, 2, None, True),
-    ([1, 2], 2, 3, None, False),
-    ([1, 2], 3, 2, None, False),
-    ({'foo'}, 1, None, None, True),
-    ({'foo'}, None, 'foo', None, True),
-    ({'foo'}, None, None, {'foo'}, True),
-    (42, None, None, 14, False),
-    (42, None, None, 42, True),
+@pytest.mark.parametrize('data,match', [
+    (dict(attribute=[], length=1), False),
+    (dict(attribute=[], data=1), False),
+    (dict(attribute=[], value=[]), True),
+    (dict(attribute=[1, 2], length=2, data=2), True),
+    (dict(attribute=[1, 2], length=2, data=3), False),
+    (dict(attribute=[1, 2], length=3, data=2), False),
+    (dict(attribute={'foo'}, length=1), True),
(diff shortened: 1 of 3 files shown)


Peque:
@ocaballeror Travis fails due to the linger test failing... :joy: (waiting for #260 to fix that). :stuck_out_tongue_winking_eye:

Peque:
@ocaballeror AppVeyor keeps failing on `ns.shutdown()` on test teardown. I try to update #259 with new occurrences. I say we fix all these CI issues before adding new features in this 0.7.0 cycle.

ocaballeror:
Yes! I never liked the arbitrary `time.sleep()`.

Let me test it a few times in Travis to make sure it works, you have my approval for everything else.

ocaballeror:
This test looks like it could benefit from this as well: https://ci.appveyor.com/project/ocaballeror/osbrain/build/1.0.42/job/xbsqth4u9m1597s4#L761

Should we change it and include it in this PR?

osbrain/tests/test_helper.py
```diff
@@ -118,25 +118,44 @@ def test_logger_received_count(agent_logger):
...
+    (dict(attribute={'foo'}, data='foo'), True),
+    (dict(attribute={'foo'}, value={'foo'}), True),
+    (dict(attribute=42, value=14), False),
+    (dict(attribute=42, value=42), True),
+    (dict(attribute=[1, 2, 3], endswith=[2, 3]), True),
+    (dict(attribute=[1, 2, 3], endswith=[3, 3]), False),
+    (dict(attribute='abcde', endswith='cde'), True),
+    (dict(attribute='abcde', endswith='ade'), False),
+], ids=[
+    'List length missmatch',
```
ocaballeror: 'Mismatch'

ocaballeror: 'mismatch'

osbrain/tests/test_helper.py
```diff
@@ -118,25 +118,44 @@ def test_logger_received_count(agent_logger):
...
+    'Data not in list',
+    'List exact value',
+    'Data in list and list length',
+    'Data not in list with length match',
+    'Data in list with length missmatch',
```
ocaballeror: 'mismatch'

osbrain/tests/test_helper.py
```diff
@@ -118,25 +118,44 @@ def test_logger_received_count(agent_logger):
...
+    'Dictionary attribute length',
+    'Key in dictionary',
+    'Dictionary exact value',
+    'Integer value missmatch',
```
ocaballeror: 'mismatch'

Peque:
Good catch with the typos! ^^

About the error in `osbrain/tests/test_agent_async_requests.py`, I had a look at it and I think it should be a separate issue. It is a good idea to change that, but if we do I'd rather re-think the whole `osbrain/tests/test_agent_async_requests.py` file, not just that particular test. So I say we merge this as-is if there are no other objections.

PR #264, opened 2018-03-14
Fix annoying `ns.shutdown()` errors and non-boxed blocks

### LINKED ISSUE #259: Windows tests seem to fail more often

Peque:
Fixes #259.

### DIFF
--- a/osbrain/proxy.py
+++ b/osbrain/proxy.py
@@ -6,6 +6,7 @@ import sys
 import time
 
 import Pyro4
+from Pyro4.errors import ConnectionClosedError
 from Pyro4.errors import NamingError
 from Pyro4.message import FLAGS_ONEWAY
 
@@ -413,4 +414,7 @@ class NSProxy(Pyro4.core.Proxy):
             Timeout, in seconds, to wait for the agents to shutdown.
         """
         self.shutdown_agents(timeout)
-        super()._pyroInvoke('daemon_shutdown', (), {}, flags=0)
+        try:
+            super()._pyroInvoke('daemon_shutdown', (), {}, flags=0)
+        except ConnectionClosedError:
+            pass
--- a/osbrain/tests/test_agent.py
+++ b/osbrain/tests/test_agent.py
@@ -276,7 +276,7 @@ def test_bind_tcp_addr_specific_port(nsproxy):
     (0, 1, False),
     (-1, 1, True),
 ])
-def test_linger(nsproxy, linger, sleep_time, should_receive):
+def test_linger(monkeypatch, nsproxy, linger, sleep_time, should_receive):
     '''
     Test linger works when closing the sockets of an agent.
     '''
@@ -284,7 +284,7 @@ def test_linger(nsproxy, linger, sleep_time, should_receive):
         def on_init(self):
             self.received = []
 
-    osbrain.config['LINGER'] = linger
+    monkeypatch.setitem(osbrain.config, 'LINGER', linger)
 
     puller = run_agent('puller', base=AgentTest)
     pusher = run_agent('pusher', base=AgentTest)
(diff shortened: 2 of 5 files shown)


Peque:
@ocaballeror Launched the test suite 2 times in AppVeyor and it seems all tests pass now.

ocaballeror:
Could be a solution. After all, the error we're getting from all these tests that fail at shutdown always has something to do with the nameserver being inaccesible and not receiving a reply from it.

We would have to sort out #265 first, though.

Peque:
@ocaballeror Let us wait for AppVeyor, but this looks good. :blush:

ocaballeror:
@Peque Can you explain why you use `monkeypatch`? I don't see what the problem is in using `osbrain.config` directly.

osbrain/proxy.py
```diff
@@ -413,4 +414,7 @@ def shutdown(self, timeout=10.):
             Timeout, in seconds, to wait for the agents to shutdown.
         """
         self.shutdown_agents(timeout)
-        super()._pyroInvoke('daemon_shutdown', (), {}, flags=0)
+        try:
+            super()._pyroInvoke('daemon_shutdown', (), {}, flags=0)
```
ocaballeror: Aside from catching the exception, maybe we could also use `Pyro4.FLAGS_ONEWAY` here. What do you think?

Peque: It would change the current behavior: to block if the name server is busy and until it actually shuts down. We could consider this if we find out it is a better default behavior, but for now I think it is better if we do not change that.

Peque:
@ocaballeror Unless `--boxed` is used (which is unmaintained now, so we should try to avoid it), that could lead to other tests reading an unexpected value of `osbrain.config`. That was actually happening and that is why the test suite could get stuck if `--boxed` was removed.

PR #270, opened 2018-03-23
Better CI configuration

Peque:
- Integrate flake8.
- Improve configuration files for CI services.


Peque:
Just pushed a fixup to recover colored output in AppVeyor.

ocaballeror:
Let's see if I understood this correctly.
You unified the list of testing dependencies between `tox.ini` and `requirements.txt`, and removed tox from it. Now users are supposed to install it manually, hence the modification on `.appveyor.yml`. Correct?

If that's the case, maybe we should mention it in the documentation, don't you think?

Peque:
I basically moved all dependencies to `setup.py`. Just like before there where some `docs` dependencies declared, now there are some `test` dependencies declared too. It is a better approach in my opinion. `tox` is actually not a required dependency for running/passing the tests locally. It is simply very convenient.

I just pushed a commit updating the documentation.

PR #271, opened 2018-03-23
Support for parallel test coverage

### LINKED ISSUE #173: Support for parallel test coverage
The idea is to improve coverage reports that sum-up results from different platforms (i.e.: Windows branches, Python < 3.5 branches...). This way we could remove `if sys.version_info` from `.coveragerc`.

Can we do it with Codecov? If we cannot, check [Code Climate](https://docs.codeclimate.com/docs/setting-up-test-coverage?utm_source=up&utm_medium=email&utm_campaign=201706&__s=cxfq3cvop3k94ucncqfc) or other alternatives.

Peque:
Depends on #270.

Fixes #173.


Peque:
@ocaballeror It seems GitHub did not update properly this PR (or at least not as I would have expected). (0f1a06be8ad47a578a42423d6c264428598b7fbc, 1d621e54c85405438f2cfd5a047ceb5c960a2df8, 0363f0a462c2c33ae5e36a83009e2e07f5989fb3) are already in master, so this PR is actually only the last commit of the 4: 0d481985ec97a45d96b05cee920e4de43e8add9a (really small one)

PR #272, opened 2018-03-23
Avoid email notifications

Peque:
Depends on #271.

I kind of got tired of those email notifications... What you think?


ocaballeror:
Thank you. I got tired of them from the moment we started integrating Appveyor.

Just one question, though, why so many commits?

Peque:
@ocaballeror As I mentioned, this depends on #271 (it includes all commits in that PR). Note that #271 depends on #270 too, which means you might want to review #270 first, #271 then and lastly this PR.

Otherwise just review this PR, but take into account it is the sum of 3 PRs.

Peque:
@ocaballeror Here again, the only relevant commit is the last one, all the others are already in `master`.

It seems I am unable to change the "base" of this PR to the current `master`. And I do not want to close and reopen the PR just for that... :joy:

ocaballeror:
The tests failed in travis, though. The error is not related to these commits at all, but still worth noticing nonetheless.

Peque:
Yeah, I noticed. It is the already familiar `ns.shutdown()` error... :sweat_smile:

ocaballeror:
@Peque Not this time. In this case it failed on `agent.shutdown()`, which is kinda weird.

PR #273, opened 2018-03-26
Use `wait_agent_attr` for async request tests

### LINKED ISSUE #269: Use `wait_agent_attr` in test_agent_async_requests.py
As mentioned here: https://github.com/opensistemas-hub/osbrain/pull/262#issuecomment-375357874

ocaballeror:
Fixes #269

### DIFF
--- a/osbrain/tests/test_agent_async_requests.py
+++ b/osbrain/tests/test_agent_async_requests.py
@@ -3,13 +3,12 @@ Test file for asynchronous requests.
 """
 import time
 
-import pytest
-
 from osbrain import run_agent
 from osbrain import Agent
 from osbrain import run_logger
 from osbrain.helper import sync_agent_logger
 from osbrain.helper import logger_received
+from osbrain.helper import wait_agent_attr
 
 from common import nsproxy  # noqa: F401
 from common import append_received
@@ -22,6 +21,7 @@ def on_error(agent):
 class Server(Agent):
     def on_init(self):
         self.received = []
+        self.blocked = False
 
 
 class Client(Agent):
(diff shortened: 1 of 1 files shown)


ocaballeror:
We definitely need to fix those `ns.shutdown()` tests

osbrain/tests/test_agent_async_requests.py
```diff
@@ -139,12 +135,12 @@ def test_unknown(nsproxy, server_client_late_reply_return):
...
-    time.sleep(1.1)
-    assert server.get_attr('received') == ['foo']
-    assert client.get_attr('received') == []
+    assert wait_agent_attr(server, data='foo')
+    assert wait_agent_attr(client, value=[])
     assert logger_received(logger,
                            log_name='log_history_warning',
-                           message='Received response for an unknown request!')
+                           message='Received response for an unknown request!',
+                           timeout=3)
```
Peque: Maybe it is not necessary to use `timetout=3` here. Can't we use the default?

osbrain/tests/test_agent_async_requests.py
```diff
@@ -75,17 +76,13 @@ def test_return(nsproxy, server_client_late_reply_return):
...
     # Wait for client to receive reply asynchronously
-    time.sleep(1)
-    assert server.get_attr('received') == ['foo', 'bar']
-    assert client.get_attr('received') == ['xfoo']
-    time.sleep(1)
-    assert server.get_attr('received') == ['foo', 'bar']
-    assert client.get_attr('received') == ['xfoo', 'xbar']
+    assert wait_agent_attr(client, data='xfoo')
+    assert wait_agent_attr(server, data='bar')
+    assert wait_agent_attr(client, data='xbar')
```
Peque: Instead of `data=...` use `value=...`. The attribute is well known and not too long. Also, order in which those messages are received matters, so `data=...` is not as reliable.

The same applies for all `data=...` in this PR.

osbrain/tests/test_agent_async_requests.py
```diff
@@ -159,16 +155,15 @@ def test_wait(nsproxy, server_client_late_reply_delay):
...
+    assert wait_agent_attr(server, data=1)
+    assert wait_agent_attr(client, data='x1')
 
     # Response not received in time
     client.send('async', 2, wait=1)
     assert logger_received(logger,
                            log_name='log_history_warning',
                            message='not receive req',
-                           timeout=1.1)
+                           timeout=3)
```
Peque: Here we are really testing the client waits for 1 second, not 3, so I do not think we should increase this timeout.

osbrain/tests/test_agent_async_requests.py
ocaballeror: It was failing with the default, hence the first fixup I pushed.

osbrain/tests/test_agent_async_requests.py
ocaballeror: What's the difference, then? Both `data` and `value` are correct in my opinion, but I figured `data` was a bit more clear, since it only includes the new value we are expecting to find, rather than the entire list with previously known values.

But now that you mention the order thing, why not change `data` to `endswith`? Would that be acceptable to you?

Peque: Yeah, that should be fine.

ocaballeror:
Let's wait until #264 is merged and then rebase this on top of it to see if the tests pass.

osbrain/tests/test_agent_async_requests.py
```diff
@@ -22,114 +23,113 @@ def on_error(agent):
...
-    client = run_agent('client', base=Client)
+    return async_client_server(late_reply)
 
-    addr = server.bind('ASYNC_REP', alias='replier', handler=late_reply)
-    client.connect(addr, alias='async', handler=append_received)
 
-    return (server, client)
+def blocked_reply(agent, request):
+    agent.received.append(request)
+    agent.blocked = True
```
Peque: Even if very unlikely, I would put `agent.blocked = True` before the append to avoid failing due to race conditions.

In example: in tests you set `blocked=False` just after checking the `received` attribute.

osbrain/tests/test_agent_async_requests.py
```diff
@@ -22,114 +23,113 @@ def on_error(agent):
...
-    time.sleep(1)
-    assert server.get_attr('received') == ['foo', 'yfoo', 'bar', 'ybar']
-    assert client.get_attr('received') == ['xfoo', 'xbar']
+    server2.unsafe.set_attr(blocked=False)
+    assert wait_agent_attr(client, value=['xbar'], timeout=.1)
+    assert server2.get_attr('received') == ['bar', 'ybar']
+
+    server1.unsafe.set_attr(blocked=False)
+    assert wait_agent_attr(client, value=['xbar', 'xfoo'], timeout=.1)
+    assert server1.get_attr('received') == ['foo', 'yfoo']
```
Peque: Use `wait_agent_attr()` instead of `server.get_attr('received')`. For consistency and also to avoid (very unlikely) race conditions.

osbrain/tests/test_agent_async_requests.py
```diff
@@ -139,8 +139,8 @@ def test_unknown(nsproxy, server_client_late_reply_return):
 
     # Manually remove the pending request before it is received
     client.set_attr(_pending_requests={})
-    time.sleep(1.1)
-    assert server.get_attr('received') == ['foo']
+    server.unsafe.set_attr(blocked=False)
+    assert wait_agent_attr(server, value=['foo'], timeout=.1)
     assert client.get_attr('received') == []
```
Peque: I would rather use `not wait_agent_attr(client, length=1, timeout=.1)`. You can also put the logger assertion before the client assertion.

Maybe you can move the `assert wait_agent_attr(server...)` up just after the `client.send()`.

ocaballeror: If you don't mind, I like the order as it is. The code looks better with `logger_received` at the end IMO.

+1 to the `not wait_agent_atttr(length=1)` thing, though. I think I'll use it in other places as well.

osbrain/tests/test_agent_async_requests.py
```diff
@@ -22,114 +23,110 @@ def on_error(agent):
...
 
-@pytest.fixture(scope='function')
-def server_client_late_reply_return():
-    def late_reply(agent, request):
-        agent.received.append(request)
-        time.sleep(1)
-        return 'x' + request
-
+def async_server_client(server_handler=append_received,
+                        client_handler=append_received):
```
Peque: You are not using a different `client_handler` anywhere, so I'd remove it from this function for simplicity.

osbrain/tests/test_agent_async_requests.py
```diff
@@ -22,114 +23,110 @@ def on_error(agent):
...
+
+    server2 = run_agent('server2', base=Server)
+    addr2 = server2.bind('ASYNC_REP', alias='replier2', handler=server_handler)
+    client.connect(addr2, alias='async2', handler=client_handler)
+
+    return (server1, server2, client)
+
+
 @pytest.fixture(scope='function')
 def server_client_late_reply_delay():
```
Peque: This fixture is no longer used, right?

osbrain/tests/test_agent_async_requests.py
Peque: Also, you always pass as first parameter the `blocked_reply` handler, except for the `yield` test. So I guess that should be the default server handler instead of `append_received`.

Peque:
PS: I do not like the `async_server_client` vs `async_server_clients` very much. The latter is for a single client and double server setup, but says "clients". And the names are too similar. Maybe `async_server` and `async_server_pair`? Or whatever else you come up with. :blush:

osbrain/tests/test_agent_async_requests.py
ocaballeror: You're right

osbrain/tests/test_agent_async_requests.py
ocaballeror: Yeah, I was trying to make this as generic as possible when I first wrote it, but this makes more sense now :+1:

ocaballeror:
@Peque yeah, the naming was temporary. I just wrote `client_servers` while I was figuring out if the idea would work at all. :smile:

osbrain/tests/test_agent_async_requests.py
```diff
@@ -139,63 +124,73 @@ def test_unknown(nsproxy, server_client_late_reply_return):
...
     # Response received in time
-    client.send('async', 1, wait=2)
-    time.sleep(1.1)
-    assert server.get_attr('received') == [1]
-    assert client.get_attr('received') == ['x1']
+    client.send('async', '1', wait=2)
+    time.sleep(1)
+    server.unsafe.set_attr(blocked=False)
+    assert wait_agent_attr(server, value=['1'], timeout=.1)
+    assert wait_agent_attr(client, value=['x1'], timeout=.1)
```
Peque: Maybe we could:

```python
client.send(...)
assert wait_agent_attr(server, ...)
server.unsafe.set_attr(...)
assert wait_agent_attr(client, ...)
```

That way we remove the sleep all together and make the test a bit faster.

osbrain/tests/test_agent_async_requests.py
```diff
@@ -139,63 +124,73 @@ def test_unknown(nsproxy, server_client_late_reply_return):
...
     logger = run_logger('logger')
     client.set_logger(logger)
     sync_agent_logger(client, logger)
 
     # Response received in time
-    client.send('async', 1, wait=2)
-    time.sleep(1.1)
-    assert server.get_attr('received') == [1]
-    assert client.get_attr('received') == ['x1']
+    client.send('async', '1', wait=2)
```
Peque: Just for consistency with other tests should we send `'foo'` instead of `'1'`? A number was previously sent to set the server blocking duration, but this is no longer the case.

If you agree, change it in the tests bellow as well. :wink:

osbrain/tests/test_agent_async_requests.py
```diff
@@ -139,63 +124,73 @@ def test_unknown(nsproxy, server_client_late_reply_return):
...
+def test_wait_timeout(nsproxy):
+    """
+    Asynchronous request-reply pattern when exceeding the maximum wait.
+    """
+    client, server = async_client_server()
+    logger = run_logger('logger')
+    client.set_logger(logger)
+    sync_agent_logger(client, logger)
+
+    client.send('async', '1', wait=1)
```
Peque: Maybe we could reduce the `wait` to 0.5 to make the test a tiny bit faster? If so, do the same with `test_wait_on_error`.

osbrain/tests/test_agent_async_requests.py
```diff
@@ -139,63 +124,73 @@ def test_unknown(nsproxy, server_client_late_reply_return):
...
     assert logger_received(logger,
                            log_name='log_history_warning',
-                           message='Received response for an unknown request!',
-                           timeout=2)
+                           message='Received response for an unknown request!')
     assert len(client.get_attr('_pending_requests')) == 0
-    assert server.get_attr('received') == [1, 2]
-    assert client.get_attr('received') == ['x1']
+    assert server.get_attr('received') == ['1']
+    assert client.get_attr('received') == []
```
Peque: I think the `assert len() == 0` should go right after `logger_received(..., 'not receive req', ...)`.

osbrain/tests/test_agent_async_requests.py
```diff
@@ -139,63 +124,73 @@ def test_unknown(nsproxy, server_client_late_reply_return):
...
     # Response received in time
-    client.send('async', 1, wait=2)
-    time.sleep(1.1)
-    assert server.get_attr('received') == [1]
-    assert client.get_attr('received') == ['x1']
+    client.send('async', '1', wait=2)
+    time.sleep(1)
+    server.unsafe.set_attr(blocked=False)
+    assert wait_agent_attr(server, value=['1'], timeout=.1)
+    assert wait_agent_attr(client, value=['x1'], timeout=.1)
```
Peque: Maybe we could remove this `# Response received in time` section from `test_wait_on_error`. It is the same code that we now have in a separate test `test_wait_in_time`.

ocaballeror: Yeah, I don't why it was there in the first place :smile:

osbrain/tests/test_agent_async_requests.py
Peque: Okay about removing the timeouts.

Reordering is important to avoid (very unlikely) situation:

```python
client.send(...)
server.unsafe.set_attr(blocked=False)  # The server did not receive the message yet,
                                       # so it will never un-block?
```

Peque:
@ocaballeror Hmmmm... About that AppVeyor failure maybe we should:

```python
client.send('async', 'foo')
assert wait_agent_attr(server, value=['foo'], timeout=.1)

# Manually remove the pending request before it is received
client.set_attr(_pending_requests={})
server.unsafe.set_attr(blocked=False)

# Client should log a warning and do not process the received message
assert logger_received(logger, ...)
assert len(client received) == 0
```

ocaballeror:
@Peque yeah, that's what I was thinking. I'll have to check if it works well.

Peque:
Looks good to me. Not sure if that `timeout=2` in your last fixup is necessary though.

If you look the coverage report, we are missing a branch. We can fix that if we edit the `test_return` adding 2 requests to a single server:

```python
client.send('async', 'foo')
client.send('async', 'foo2')
client.send('async2', 'bar')

# ...
server1.unsafe.set_attr(blocked=False)
assert wait_agent_attr(client, value=['xbar', 'xfoo', 'xfoo2'], timeout=.1)
```

What you think?

ocaballeror:
@Peque I don't understand. How is that supposed to affect the coverage if we are just calling the same functions but twice?

Peque:
@ocaballeror First call there is no connection, so one is created. Second time there is connection, so no need to re-create it. Thus the branching.

ocaballeror:
According to Google, 'qux' is the next word in the 'foo' 'bar' 'baz' sequence, so I'll go with that :smile:

PR #275, opened 2018-03-27
Some more linting

### LINKED ISSUE #237: Update style/lint checkers
https://github.com/PyCQA

Peque:
Fixes #237.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,8 @@ matrix:
   include:
     - python: 3.6
       env: TOXENV=docs
+    - python: 3.6
+      env: TOXENV=lint
 
 install:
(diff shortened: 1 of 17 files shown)


ocaballeror:
The linting and docs checks are not run on appveyor. Are you aware of that?

Peque:
Yeah, and they are only run once now, rather than once for each Python version. Linting is expected to give the same results for any Python version and for the documentation generation we do not really need to support all Python versions (although they should be supported anyway unless something is broken on Sphinx end).

Thought it was a good idea to separate them. Easier to see if your tests are failing due to styling/linting or due to incompatibilities with a Python version or due to broken tests.

PR #281, opened 2018-04-03
Ignore async replies after maximum wait

### LINKED ISSUE #280: Agents receive async replies even after maximum `wait`
After the maximum wait time expires, the `on_error` handler is executed correctly, but agents still receive the reply and will process it as usual, which is not the expected behavior.

The request should be removed from `_pending_requests` after the timeout has expired so agents ignore the replies if they come in too late.

ocaballeror:
Fixes #280

### DIFF
--- a/osbrain/tests/test_agent_async_requests.py
+++ b/osbrain/tests/test_agent_async_requests.py
@@ -165,10 +165,17 @@ def test_wait(nsproxy, server_client_late_reply_delay):
 
     # Response not received in time
     client.send('async', 2, wait=1)
+    assert len(client.get_attr('_pending_requests')) == 1
     assert logger_received(logger,
                            log_name='log_history_warning',
                            message='not receive req',
                            timeout=1.1)
+
+    assert logger_received(logger,
+                           log_name='log_history_warning',
+                           message='Received response for an unknown request!',
+                           timeout=2)
+    assert len(client.get_attr('_pending_requests')) == 0
     assert server.get_attr('received') == [1, 2]
     assert client.get_attr('received') == ['x1']
 
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -1602,6 +1602,8 @@ class Agent():
         """
         if uuid not in self._pending_requests:
             return
+
+        del self._pending_requests[uuid]
         if not on_error:
             warning = 'Did not receive request {} after {} seconds'.format(
                 uuid, wait)


ocaballeror:
@Peque you're right. The code was actually wrong, since `del` was not being called with the default `on_error` :sweat_smile:

osbrain/tests/test_agent_async_requests.py
```diff
@@ -169,6 +169,7 @@ def test_wait(nsproxy, server_client_late_reply_delay):
                            log_name='log_history_warning',
                            message='not receive req',
                            timeout=1.1)
+    time.sleep(1)
     assert server.get_attr('received') == [1, 2]
     assert client.get_attr('received') == ['x1']
```
Peque: The test is not very robust in my opinion. If you do `time.sleep(0.9)` it passes too, so that timing is a bit too close to the limit value.

Maybe you should increase the sleep. Alternatively (better in my opinion) you can use `logger_received` again, as the agent should log a warning when it receives a message for an unknown pending request.

Also, you could `assert` the value of `_pending_requests`. You could verify that:

- After sending the request, the attribute has one pending request.
- After the `wait` (once the logger gets the first warning), the attribute is empty.
- If you want to go further (maybe it is not necessary), you can check that the request UUID is part of both logged messages. Note that `logger_received` accepts regular expressions.

Also, we should probably split this test in two: `test_wait_in_time()`, `test_wait_timeout()`.

ocaballeror: I agree with most of those things, but I think we should probably wait for the refactoring in #273 before making any of those changes.

If you're OK with it, I'll add the assertions on `_pending_requests` to make this test more robust, but hold on with the splitting in two, and consider adding that to the other PR. What do you think?

Peque: If you want to wait and merge this before splitting the test in two, that is fine for me.

I think the other changes should be included though:

- Using `logger_received` to check the unknown pending request warning and avoid the `time.sleep()`.
- Checking the value of `_pending_requests` (at least the size of the attribute). Both after sending the request and after the wait.

After all, this PR is (should be) independent from #273.

ocaballeror: I could also use `wait_agent_attr` and solve both problems. Take a look at the last fixup I pushed.

Peque: I prefer explicitness in tests:

- I would say `assert len() == 1` (exactly one).
- I would use separate `logger_received` and  `assert len() == 0`. This way we test 2 things:
  - That the agent actually received the reply at some point and logged a warning message, which is useful.
  - The length of the attribute.

In tests clarity and precision should be over LOC count and neat/smart tricks that do the job.

Peque:
@ocaballeror I think I commented on an outdated commit in my last review, but the comment still applies. :sweat_smile:

osbrain/tests/test_agent_async_requests.py
ocaballeror: Got it :+1:

PR #286, opened 2018-04-03
Check homogeneous quoting style in source code

Peque:
Found this and thought... why not? :joy:

### DIFF
--- a/examples/explicit_serialization.py
+++ b/examples/explicit_serialization.py
@@ -1,7 +1,7 @@
-'''
+"""
 A simple example in which the different serialization setting
 options are shown.
-'''
+"""
 import osbrain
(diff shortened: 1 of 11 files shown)


ocaballeror:
I'm guessing this is some sort of standard? Or is it just because you thought it looks better?

Peque:
Yeah, PEP8 recommends PEP257 for docstrings, wich recommends triple double quotes. This is also what we were using (almost) everywhere.

For string quotes I do not know if there is an official recommendation. I think the recommendation is to simply stick to one or another, for consistency. As you can see we were already using single string quotes (almost) everywhere. I also think this is the *de facto* standard and most common approach.

PR #289, opened 2018-04-05
Test nameserver shutdown with a ConnectionClosedError

### LINKED ISSUE #279: Uncovered lines
This exception handling in `osbrain/proxy.py` is not always covered by the test suite:

```python
class Proxy():
    def shutdown(self, timeout=10.):
        # ...
        except ConnectionClosedError:
            pass
```

We should create a test that forces a `ConnnectionClosedError` on shutdown.

ocaballeror:
Fixes #279

### DIFF
--- a/osbrain/tests/test_proxy.py
+++ b/osbrain/tests/test_proxy.py
@@ -7,10 +7,14 @@ import pickle
 import pytest
 from threading import Timer
 
+from Pyro4.errors import ConnectionClosedError
+
 import osbrain
 from osbrain import run_agent
+from osbrain import run_nameserver
 from osbrain import Agent
 from osbrain import AgentProcess
+from osbrain import NameServer
 from osbrain import Proxy
 from osbrain.proxy import locate_ns
 from osbrain.helper import wait_agent_attr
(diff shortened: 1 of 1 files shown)


osbrain/tests/test_proxy.py
```diff
@@ -325,3 +329,13 @@ def shoot(self):
     assert not wayne._next_oneway
 
     assert wait_agent_attr(wayne, value=20 * ['bang!'], timeout=1.5)
+
+
+def test_agent_proxy_shutdown_connectionclosed():
```
Peque: `test_nameserver_proxy_*` instead?

Peque: Missing test docstring. Even if just something like "ConnectionClosedError exceptions are allowed when shutting down a name server from a proxy".

ocaballeror: Copy-pasting is a hard thing to do well...

Peque: Or even better: "... when remotely shutting down the name server daemon from a proxy".

Or whatever you come up with. Just something a bit descriptive. :joy:

Peque: I know how you feel. :joy:

Issue #255, opened 2018-02-27
Add linger parameter to `.close()` and `.close_all()`


ocaballeror:
As mentioned in https://github.com/opensistemas-hub/osbrain/pull/252#issuecomment-368051072

This is mostly for Windows compatibility, since it won't support changing the linger value in the global config dictionary.

By the way, this also needs tests for both methods.

Peque:
Yeah, let us not work on this issue for now. Maybe we can fix the global configuration problem instead.

Peque:
There is #256 now, so i am moving this to milestone 0.8.0.

Note that this change would probably require not only adding more tests, but also allowing to pass a `linger` parameter to other method calls such as `.shutdown()` (and more tests, of course). So I am not sure it is worth it.

Anyway, let us discuss this for 0.8.0 instead. :stuck_out_tongue_winking_eye:

PR #294, opened 2018-04-06
Safety

### LINKED ISSUE #287: Socket file removal tests randomly failing on Travis
https://travis-ci.org/opensistemas-hub/osbrain/jobs/362494498
https://travis-ci.org/ocaballeror/osbrain/jobs/362021572
https://travis-ci.org/ocaballeror/osbrain/jobs/362012690

Occurrences
-----------------

Occurrences | Test
--- | ---
1 | `test_agent_close_ipc_socket_agent_blocked_nameserver_shutdown`
1 | `test_agent_close_ipc_socket_agent_crash_nameserver_shutdown`
3 | `test_agent_close_ipc_socket_agent_kill`

Peque:
Fixes #287.

Not really proud about the final result, but should definitely be a step forward...

### DIFF
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -1770,8 +1770,7 @@ class Agent():
         """
         Close a socket using the provided linger value.
         """
-        if linger is None:
-            linger = get_linger()
+        linger = get_linger(linger)
         socket.close(linger=linger)
 
     def close(self, alias, linger=None):
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -259,17 +259,6 @@ class Agent():
         self._context = zmq.Context()
         self._poller = zmq.Poller()
 
-        # A loopback socket where, for example, timers are processed
-        self.bind('REP', alias='loopback', addr='loopback',
-                  handler=self._handle_loopback, transport='inproc',
-                  serializer='pickle')
-
-        # This in-process socket handles safe access to
-        # memory from other threads (i.e. when using Pyro proxies).
-        self.bind('REP', alias='_loopback_safe', addr='_loopback_safe',
-                  handler=self._handle_loopback_safe, transport='inproc',
-                  serializer='pickle')
-
         self._set_attributes(attributes)
         self.on_init()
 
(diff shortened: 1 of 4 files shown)


osbrain/agent.py
```diff
@@ -1673,36 +1686,42 @@ def run(self):
...
+        If the agent is running the ZMQ context is terminated to allow the
+        main thread to quit and do the tear down.
+        """
+        self.stop_all_timers()
+        if self._running:
+            self._context.term()
+        else:
+            self._die()
+
+    def _die(self):
```
ocaballeror: The `_die` method seems overly complicated to me. I think it just adds an unnecessary amount of code. Go on with whichever solution you like best, but consider this implementation as well:
```py
def run(self):
    ...
    if self._kill_now:
        self.kill()

def kill(self):
    self.stop_all_timers()
    if self._running:
        self._context.term()
    else:
        self.close_all()
        self._pyroDaemon.shutdown()

# And delete the `die` method.
```

If you end up keeping yours, maybe consider renaming `_kill_now` to `_die_now`

osbrain/agent.py
```diff
@@ -1743,41 +1762,44 @@ def has_socket(self, alias):
...
+    def _delete_socket_entries(self, entries):
+        for entry in entries:
+            del self._socket[entry]
+
+    def close(self, alias, linger=None):
         """
         Close a socket given its alias and clear its entry from the
         `Agent.socket` dictionary.
         """
+        if linger is None:
```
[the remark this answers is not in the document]
Peque: As mentioned in my other comment it is to set linger to zero when killing (i.e.: killing should be fast).

This would definitely need more tests. I do not think we need to create a new issue, as there is #255 already.

Anyway I think you are right. I will either add some tests or undo that linger change (and remove the `_die()` method, which would make little sense then.

osbrain/agent.py
```diff
@@ -1709,9 +1709,9 @@ def kill(self):
         if self._running:
             self._context.term()
         else:
-            self._die()
+            self._die(linger=0)
 
-    def _die(self):
+    def _die(self, linger):
```
ocaballeror: Not really necessary, but I think it looks better 😁 

Consider changing the method signature to:
```py
def _die(self, linger=None)
```
And then change the end of `.run()` to `self._die()`

osbrain/agent.py
ocaballeror: I see what you mean. I don't want to complain too much about this since it makes sense for the most part, so check out my latest revision and then we can move on with this.

osbrain/agent.py
Peque: Changed signature and renamed attribute.

ocaballeror:
Maybe merge #296 first? That way we can merge this with the tests passing.

Peque:
@ocaballeror Rebased, squashed and added socket close with linger tests.

Peque:
Also pushed a new commit to fix: https://travis-ci.org/opensistemas-hub/osbrain/jobs/364043140

I looked for other possible occurrences in our test suite and that seems to be the last one.

PR #296, opened 2018-04-09
Fix agent inheritance test to wait for running

Peque:
To prevent this from happening again:

https://travis-ci.org/opensistemas-hub/osbrain/jobs/364022226


Peque:
Yeah, rather than started I think it just happened (i.e.: if we run our test suite another 100 times, maybe it would not appear again). :sweat_smile:

PR #302, opened 2018-07-23
Avoid using version 17 of PyZMQ for now

Peque:
See https://github.com/zeromq/pyzmq/issues/1170.

The test suite [breaks](https://travis-ci.org/opensistemas-hub/osbrain/builds/405590284) with PyZMQ v17 while [passes](https://travis-ci.org/opensistemas-hub/osbrain/builds/407155149) when `pyzmq` is set to <v17 (v16).

### DIFF
--- a/osbrain/__init__.py
+++ b/osbrain/__init__.py
@@ -39,7 +39,7 @@ else:
     config['IPC_DIR'].mkdir(exist_ok=True, parents=True)
 
 
-__version__ = '0.6.1'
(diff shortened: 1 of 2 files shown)


ocaballeror:
#303

Everything titled "for now" probably needs a new issue, so it doesn't become "for ever".

Peque:
@ocaballeror I'd say "surely" instead of "probably". :stuck_out_tongue_winking_eye:

PR #304, opened 2018-07-30
remove b' from ipc socket names

minrk:
`str(b"x")` is `b"x"` so ipc sockets were being created as `~/.osbrain_ipc/b'37a76fda9a264...'`. Instead, cast bytes to str via the `bytes.decode()` method.


Peque:
@minrk Is it just an aesthetics change or did you find any issue with this?

@ocaballeror What do you say? I think it is fine to merge this; there is no practical difference and maybe socket file names look better without the `b''`.

PR #305, opened 2018-07-30
fix tests with pyzmq 17

### LINKED ISSUE #303: Update PyZMQ to version 17
As pointed in #302, it doesn't work right now, so we should probably find a way to properly use the new version, rather than avoiding it forever.

Might be a PyZMQ issue. Reference: https://github.com/zeromq/pyzmq/issues/1170.

See also: https://github.com/zeromq/libzmq/issues/3204

minrk:
libzmq 4.2, which is bundled with pyzmq 17 but can be used with ~any version of pyzmq, [does not cleanup ipc files](https://github.com/zeromq/libzmq/issues/2764) after closing sockets. The fix is to cleanup these files after closing the binding socket.

closes #303

### DIFF
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -1772,6 +1772,15 @@ class Agent():
         """
         linger = get_linger(linger)
         socket.close(linger=linger)
+        address = self._address[socket]
+        if address.transport == 'ipc':
+            # cleanup ipc sockets
+            try:
+                os.unlink(address.address)
+            except FileNotFoundError:
+                # someone else already removed it,
+                # e.g. libzmq < 4.2
+                pass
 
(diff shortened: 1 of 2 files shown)


Peque:
Thanks for this PR and for finding the issue with current ZMQ not unlinking those files. It is unfortunate but we definitely need to fix that on our side then.

I will review your changes and fix the remaining issues with CI. You have helped enough already! :blush:

Peque:
Closed in vafour of https://github.com/opensistemas-hub/osbrain/pull/307.

@minrk Your commits where squashed but you kept the authorship: afd59ac0a3db6f1eeed8713e40d7f5c06f9bfee0

Thanks for your help! :blush:

PR #307, opened 2018-08-08
Fix PyZMQ 17 compatibility issues

### LINKED ISSUE #303: Update PyZMQ to version 17

Peque:
By manually cleaning IPC files after closing sockets.

Fixes #303. Closes #305.

### DIFF
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -1,6 +1,7 @@
 """
 Core agent classes.
 """
+import contextlib
 from datetime import datetime
 import errno
 import inspect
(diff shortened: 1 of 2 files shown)


ocaballeror:
Is this not the same as #305 ?

Peque:
@ocaballeror Almost. Only we unlink from the binder only and we make sure we close all sockets in a channel. I do not mind implementing this over #305 if you think it is better (to keep @minrk's contribution in the history).

Peque:
@ocaballeror Yeah, did not want to bother him with rebasing, squashing or commit message style rules, but I guess we can be more lenient when accepting external/occasional contributions. :blush:

I'll do the rebase/squash/rename while keeping his authorship and then add the current changes as a "refactor". :+1:

Peque:
@ocaballeror Done.

PR #315, opened 2018-11-13
Fix linting errors

ocaballeror:
Flake8 was updated and a bunch of new errors started showing up: https://travis-ci.org/opensistemas-hub/osbrain/jobs/452151480

Apparently we weren't using `noqa` correctly either. `flake8: noqa` would ignore warnings for the entire file, regardless of whether we used them in a single line or not. This behavior [has been changed](https://gitlab.com/pycqa/flake8/merge_requests/219) in the new version, and that uncovered a couple more errors.

### DIFF
--- a/osbrain/proxy.py
+++ b/osbrain/proxy.py
@@ -208,8 +208,8 @@ class Proxy(Pyro4.core.Proxy):
         self._next_oneway = True
         return self
 
-    def _pyroInvoke(self, methodname, args, kwargs,
-                    flags=0, objectId=None):  # flake8: noqa
+    def _pyroInvoke(self, methodname, args, kwargs,  # noqa: N802, N803
+                    flags=0, objectId=None):
         """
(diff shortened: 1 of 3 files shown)


osbrain/proxy.py
```diff
@@ -246,7 +246,8 @@ def _is_safe_method(self, methodname):
                 methodname not in ('run', 'get_attr', 'kill', 'safe_call',
                                    'concurrent', 'is_running'))
 
-    def _remote_call(self, methodname, args, kwargs, flags, objectId):
+    def _remote_call(self, methodname, args, kwargs, flags,  # noqa: N803
+                     objectId):
```
Peque: Cannot we put the `  # noqa` comment after the `:`?

ocaballeror: Nope. I tried that already and it doesn't work :disappointed:

Peque:
Looks good to me. :+1:

PR #320, opened 2019-02-06
Fix linting errors

ocaballeror:
Two things:
1. `pep8-naming` was updated and now `flake8` (sadly) works as expected, which means it doesn't support using a single `noqa` to ignore the entire function definition. Now you need separate comments for every affected line. I opened [an issue](https://gitlab.com/pycqa/flake8/issues/506) but they told me they're not going to implement it and closed it immediately.

2. The `per-file-ignores` plugin was integrated into `flake8` and it's not necessary anymore.

### DIFF
--- a/osbrain/proxy.py
+++ b/osbrain/proxy.py
@@ -208,8 +208,8 @@ class Proxy(Pyro4.core.Proxy):
         self._next_oneway = True
         return self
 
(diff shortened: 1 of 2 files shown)


ocaballeror:
Also, congrats on reaching 100 stars 🎉 🎉

PR #328, opened 2019-04-11
Lower default log level in agent info messages

Peque:
@ocaballeror I think you once suggested this. It is time now! :stuck_out_tongue_winking_eye: 

Fixes https://github.com/opensistemas-hub/osbrain/issues/329

### DIFF
--- a/osbrain/agent.py
+++ b/osbrain/agent.py
@@ -464,7 +464,7 @@ class Agent():
         """
         Stop the agent. Agent will stop running.
         """
-        self.log_info('Stopping...')
+        self.log_debug('Stopping...')
         self._keep_alive = False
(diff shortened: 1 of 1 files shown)


ocaballeror:
Did I? Well, then I agree with 8-months-ago me. 

What about the environmental variable? I think that would be interesting from an end user's perspective. OSBRAIN_DEBUG maybe?

Peque:
@ocaballeror Yeah, that is a different issue. I just created https://github.com/opensistemas-hub/osbrain/issues/329 to differentiate them.

This PR will only close https://github.com/opensistemas-hub/osbrain/issues/329. Adding a variable to configure logging level would still be a wanted feature.

PR #338, opened 2019-08-01
Poller cleanup

Peque:


osbrain/tests/test_agent.py
```diff
@@ -545,6 +545,36 @@ def test_close(nsproxy):
...
 
+def test_close_socket_poller_cleanup(nsproxy):
+    """
+    Closed sockets should be cleaned-up from the poller.
+
+    When Bob connects to Alice, a new socket is registered in Bob's poller.
+    Once we close that socket, it should no longer be registered in the poller.
+    """
+    def rep_handler(agent, message):
+        return 'OK'
```
ocaballeror: You can merge the other PR and use the new echo handler to fix the coverage diff issue here.

Peque: Yeah, thanks for updating it! :blush:

Already rebased against master and added a fixup to change that.

Peque:
@ocaballeror I did not like the old test very much, but I am not against adding another test if required.

The question would be: what would you test? I would not want to test that, after closing a socket and waiting an arbitrary amount of time, the agent is still alive (in particular the arbitrary amount of time). Could you reproduce the error you were getting without that time sleep?

Otherwise, we could simply test `assert not agent.has_socket('alias')` as we do for bound sockets, but not sure it adds much.

ocaballeror:
Yes, the test would need to be changed now that we have a new one to test this specific bug.

I'm OK with the `has_socket` test, I think that should be enough.

Peque:
@ocaballeror Pushed the little test: 591b8b1fb1f5e422a5d9c33493a7324d8280fff8

If it is okay with you, I will merge this and make a minor release.

Peque:
@ocaballeror Thanks a lot! New version is already available in PyPI. :blush:

Issue #277, opened 2018-03-27
Black formatting


Peque:
When it gets out of beta: https://github.com/python/black/issues/517

PR #343, opened 2019-09-20
Black

Peque:
Using Black formatting now that we are integrating the tool in other internal projects.

@ocaballeror We all knew this day would come. :joy: 

Closes https://github.com/opensistemas-hub/osbrain/issues/277.


ocaballeror:
Yay! I love black (for the most part), I've been using it in all my projects for a few months now.

However, I'm really not a fan of the string normalization feature. It doesn't really enforce the PEP8 standard, and it forces you to use double quotes everywhere, which generates a ton of changes everywhere. Also, they look uglier, at leas to me.

I'd say we add the `skip-string-normalization` option to `pyproject.toml`

Peque:
@ocaballeror I prefer single quotes too. Wanted to be more impartial and not make that decision based on my own preferences. You may have given me an excuse to leave the quotes as they were before... :stuck_out_tongue_winking_eye: 

I will probably revert that commit tomorrow (it is separated from the other Black-formatting changes).

ocaballeror:
I know, black is meant to be used with all its defaults so that there is a consistent style across the board and you don't have to decide anything. 

However, since quotes are down to personal choice anyway, I'd rather keep OUR personal choice rather than Black's developers. 

Also, it's not like we're not enforcing a consistent quoting style in this project, we have a dedicated flake8 plugin just to ensure everything follows a standard. 

Sep 23, 2019, 23:44 by notifications@github.com:

Peque:
@ocaballeror Done.

Until there is an official recommendation (i.e.: PEP), we will stick to single-quote style. Not only to match our preferences, which is probably a weak argument, but to also minimize the number of LoC modified for the style change, since we were already being consistent in that respect.

osbrain/tests/common.py
```diff
@@ -2,17 +2,22 @@
...
-                               reason='Windows does not support IPC')
+skip_windows = mark.skipif(
+    sys.platform == 'win32', reason='Not supported on windows'
+)
+skip_windows_port_reuse = mark.skipif(
+    sys.platform == 'win32', reason='Windows allows port reuse'
+)
+skip_windows_any_port = mark.skipif(
+    sys.platform == 'win32',
+    reason='Windows allows binding to well ' 'known ports',
```
ocaballeror: You can probably join these two strings

tox.ini
```diff
@@ -29,14 +29,16 @@ commands =
         {posargs:-n 8}
 
 [testenv:lint]
-deps =
-    -e.[dev]
 commands =
+    black . --diff --check
```
ocaballeror: Maybe add a `-q` here to supress the emoji messages

tox.ini
```diff
@@ -29,14 +29,16 @@ commands =
...
     flake8
 
+[testenv:fixlint]
+commands =
+    black .
```
ocaballeror: Maybe add a `-q` here to supress the emoji messages

docs/source/conf.py
```diff
@@ -197,56 +197,64 @@
...
 # Grouping the document tree into LaTeX files. List of tuples
 # (source start file, target name, title,
 #  author, documentclass [howto, manual, or own class]).
 latex_documents = [
-  ('index', 'osBrain.tex', u'osBrain Documentation',
-   u'Open Sistemas de Información Internet S.L.', 'manual'),
+    (
+        'index',
+        'osBrain.tex',
+        u'osBrain Documentation',
```
ocaballeror: Now that we're at it, do we need the u-strings here? They look very... python2

osbrain/address.py
```diff
@@ -89,6 +89,7 @@ class AgentAddressTransport(str):
     """
     Agent's address transport class. It can be 'tcp', 'ipc' or 'inproc'.
     """
+
```
ocaballeror: I really don't like these blank lines, but apparently they are part of PEP257.

.appveyor.yml
```diff
@@ -7,7 +7,6 @@ environment:
   matrix:
     - PYTHON: "C:\\Python37-x64"
     - PYTHON: "C:\\Python36-x64"
-    - PYTHON: "C:\\Python35-x64"
```
ocaballeror: I support this move, but why was it done?

osbrain/tests/common.py
Peque: Sure, nice catch! :joy:

Fixed in 158568066e6b8279219d91ebe1913c957e407498

tox.ini
Peque: I don't mind them. :stuck_out_tongue_winking_eye:

tox.ini
Peque: I don't mind them. :stuck_out_tongue_winking_eye:

docs/source/conf.py
Peque: Probably not. I think this should be a separate issue anyways. We could update the whole `conf.py`, which is from 2014.

.appveyor.yml
Peque: It was not my plan. But it seems AppVeyor was using a very old `pip` which would not allow to interpret `black;python_version>"3.5"`. So pipelines were failing.

Tried to update `pip` from within the `.appveyor.yml` and from within `tox.ini` without success, so I decided to drop it. Feel free to try to fix it if you want to spend your time with that.

We still have Python 3.5 coverage in Travis and 3.5 will be dropped anyways next year so I did not feel too bad while dropping it... :joy: Also, AppVeyor pipelines are the slowest, and this reduces the overall CI run time.

ocaballeror: Yeah, sure, I don't think this is needed either. It's nice to have, but I don't think it's worth it to invest too much time into 3.5

